PR #11, opened 2015-07-21
Do not use activerecord classify since it de-pluralizes Meta to Metum.

graydon:


nullstyle:
Ugh, sorry for you having to wade through the ruby codegen... that was the first of the bunch and is by far the worst.  The other generators have at the very least extracted a method called `name` that is used consistently to go from "xdr name" to "language specific name".

Regardless, I would actually like to change one thing with this.  Ruby, unlike XDR hangs some semantics on the casing of a name that we need to preserve.  Namely, an identifier with an initial capital is a constant, which is invariably what we want with any xdr declaration.

The non-inflecting, non-singularizing variation of `classify` is `camelize`, which is what we actually want to use in this instance.  It will ensure that we get an identifier that is suitable for a ruby class, even if the xdr name has an initial lower case letter.

graydon:
No problem, I was pretty sure `to_s` was an overzealous "fix" but I was running low on ability-to-care at that point and willing to take anything that'd get us to a working codebase. Whatever works best.

nullstyle:
I'm going to go ahead and merge this in... I'll roll the change over to camelize in one of my upcoming PRs

PR #13, opened 2015-08-25
fix compile error of ThresholdIndices.java

codeck:
formal code generates "ThresholdIndexes(int value)"  in "enum ThresholdIndices"


nullstyle:
Thanks for the catch!

PR #16, opened 2015-11-11
Capitalize methods

bartekn:


nullstyle:
+1

This will work for now, but at some point we should refactor this to use `camelize`: http://api.rubyonrails.org/classes/ActiveSupport/Inflector.html#method-i-camelize

In this case, it would just be `m.name.camelize`

bartekn:
I tried to use `camelize` but it also makes every letter except the first lower case, so `accountID` becomes `Accountid'.

nullstyle:
There are system's to work around that, by adding custom inflections, but point taken.   I can look at it more in the future.

bartekn:
Merging current solution.

PR #17, opened 2015-11-12
Fix padding in Java

bartekn:
`pad` methods in both `XdrDataInputStream` and `XdrDataOutputStream` were wrong. For `mCount` equal 4 they should pad/skip 0 bytes but were padding/skipping 4.


nullstyle:
:+1:

PR #20, opened 2015-11-17
.NET (C#) output

raymens:
I've performed several succesfull tests, but there's still a chance that in some cases it will generate bad code.

Can generate C# code using the Stellar XDR files and is able to decode the LedgerHeader using a base64 string. The java generator is used as a source.

Output isn't optimal, there's probably still a bunch of improvements that can be made. It isn't properly indented (at all). It uses automatic property initialization that's been introduced in the latest C# version (C# 6). It's technically possible to not use this and expand the userbase if that's preferred as it's just syntactic sugar.

Sample typedef (after a manual format):

``` csharp
// Automatically generated by xdrgen 
// DO NOT EDIT or your changes may be overwritten

namespace Test.XDR
{
    // === xdr source ============================================================
    //  typedef opaque Hash[32];
    //  ===========================================================================
    public class Hash
    {
        public byte[] InnerValue { get; set; } = default(byte[]);

        public Hash() { }

        public Hash(byte[] Hash)
        {
            InnerValue = Hash;
        }

        public static void Encode(IByteWriter stream, Hash encodedHash)
        {
            int HashSize = encodedHash.InnerValue.Length;
            XdrEncoding.WriteFixOpaque(stream, (uint)HashSize, encodedHash.InnerValue);
        }

        public static Hash Decode(IByteReader stream)
        {
            Hash decodedHash = new Hash();
            int Hashsize = 32;
            decodedHash.InnerValue = XdrEncoding.ReadFixOpaque(stream, (uint)Hashsize);
            return decodedHash;
        }
    }
}
```

### DIFF
--- a/lib/xdrgen/generators.rb
+++ b/lib/xdrgen/generators.rb
@@ -6,6 +6,7 @@ module Xdrgen::Generators
   autoload :Go
   autoload :Javascript
   autoload :Java
+  autoload :Dotnet
 
   def self.for_language(language)
     const_get language.to_s.classify
--- a/lib/xdrgen/generators/dotnet.rb
+++ b/lib/xdrgen/generators/dotnet.rb
@@ -1,0 +1,638 @@
+module Xdrgen
+  module Generators
+
+    class Dotnet < Xdrgen::Generators::Base
+
+      def generate
+        render_lib
+        render_definitions(@top)
+      end
+
+      def render_lib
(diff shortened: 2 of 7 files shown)


nullstyle:
We're trying to figure out the implications of including the code from github.com/ExM/OncRpc that you've included.  I'll update this PR when I know more.

leighmcculloch:
Closing as C# was added in https://github.com/stellar/xdrgen/issues/58.

PR #21, opened 2016-01-26
Checking an upper-bound of bytes a string may contain in go

bartekn:
This PR adds support for checking an upper-bound of bytes that strings may contain, like:
```xdr
string text<28>;
```

To achieve this:
* A container for `string` with upper-bound type was changed to `[m]rune` where `m` is an upper-bound of bytes that a string may contain,
* Getters: `Get#{name arm}` and `Must#{name arm}` have a different body for `[m]rune` fields: they convert `[m]rune` and return `string`.
* Setters: `New#{name union}` have a different body for `[m]rune` fields: they convert `string` and save `[m]rune`.

The previous version of generated code did not check an upper-bound of the strings during decoding and encoding thus allowing to decode and encode objects with string fields that exceeded the maximum length specified in XDR files.

### DIFF
--- a/lib/xdrgen/ast/declarations/string.rb
+++ b/lib/xdrgen/ast/declarations/string.rb
@@ -3,9 +3,11 @@ module Xdrgen::AST::Declarations
     delegate :name, to: :identifier
 
     def size
-      size_str = size_spec.size_t.text_value
+      size_spec.size
+    end
 
-      size_str.to_i if size_str.present?
+    def fixed?
+      size_spec.is_a?(Xdrgen::AST::VarSize)
     end
   end
 end
--- a/lib/xdrgen/ast/typespecs/string.rb
+++ b/lib/xdrgen/ast/typespecs/string.rb
@@ -4,6 +4,7 @@ module Xdrgen::AST::Typespecs
     
     delegate :size, to: :decl
     delegate :name, to: :decl
+    delegate :fixed?, to: :decl
 
   end
 end
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -293,6 +293,7 @@ module Xdrgen
           package #{@namespace || "main"}
 
           import (
+            "bytes"
             "io"
             "fmt"
 
@@ -344,7 +345,11 @@ module Xdrgen
         when AST::Typespecs::Quadruple
           raise "no quadruple support for go"
         when AST::Typespecs::String
-          "string"
+          if type.fixed?
+            "[#{type.size}]rune"
+          else
+            "string"
+          end
         when AST::Typespecs::UnsignedHyper
           "uint64"
         when AST::Typespecs::UnsignedInt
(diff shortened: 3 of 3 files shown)


nullstyle:
I don't think this actually works how we want.  First off, correct me if I'm wrong, but this will involve having normal XDR structs return a `[m]rune`, whereas unions will a return a string since you've edited the accessors.  If this is the case, I don't believe that mismatch is what we want.  A consumer of `string<4>` should get the same value regardless if it is embedded in a union or a struct.

Secondly, I think this is actually solving the problem at too high of a level.  If you trace the specific problem (i.e. trusting a length attribute provided by external data) down to the lowest level, you'll see that the root is actually that the go-xdr library we use has no mechanism to restrict length based upon the xdr schema.  To me, that says the proper solution is two part:

1.  `xdrgen` needs some way to specify the length restriction as defined by the schema being used for code generation
2.  `go-xdr` needs some way of reading that length specification and using it during decode.

IMO, one solution would be to have `xdrgen` simply add struct tags to the struct definition like so:

```go
type Foo struct {
   SomeString string `xdrmaxlength="20"`
}
```
You would then have to consume those tags from within go-xdr in `decodeStruct` and `decodeUnion` as we do with the `xdropaque` tag.

----

Beyond that, I just want to make a couple of comments about the code, since I don't quite understand things.  I'll make comments inline.

lib/xdrgen/ast/declarations/string.rb
```diff
@@ -3,9 +3,11 @@ class String < Base
     delegate :name, to: :identifier
 
     def size
-      size_str = size_spec.size_t.text_value
+      size_spec.size
```
nullstyle: As far as I can tell, you are changing semantics here.  XDR strings are not required to have an upper bound (they can be declared like `string<>`)... Am I correct that you've changed that behavior?

lib/xdrgen/ast/declarations/string.rb
```diff
@@ -3,9 +3,11 @@ class String < Base
...
+    end
 
-      size_str.to_i if size_str.present?
+    def fixed?
+      size_spec.is_a?(Xdrgen::AST::VarSize)
```
nullstyle: A string is _never_ fixed size in XDR... all strings are variable sized according to the spec.

bartekn:
Sorry for a delayed response. Today I realized this PR is not finished yet and we need to fix this before it will start making problems.

I agree that using tags would be better but the problem is it won't be possible for custom types like (because you can't add tags to a _flat_ variables):
```go
type Hash [32]byte
type Uint256 [32]byte
type DataValue []byte
type String32 string
```
I see that we are not using those types directly in Stellar so we're good with using tags but xdrgen and go-xdr can be used by others too. I can see following options (from the easiest):
* Ignore checking length of flat variables (the code will still be vulnerable in some cases),
* If we decide to use fixed length types only (so no `string`) we can probably use reflection to get the max length of underlying type. On the other hand, it will make usability worse.
* Rewrite xdrgen in a way that everything is a struct. For example, `Hash` type from above will become:

  ```go
  type String32 struct {
    Value string `xdrmaxlength="32"`
  }
  ```

nullstyle:
I'm going to close this @bartekn as the replacement is here:  https://github.com/stellar/xdrgen/pull/22

Any feedback is welcome.

PR #22, opened 2016-04-13
Add sizing restrictions to generated go code

nullstyle:


nullstyle:
Some notes about this implementation:

Struct fields use the `xdrmaxsize` tag to specify their length.  Additionally, typedefs can contain sized declarations and we support this by making the typedef confirm to the `Sized` interface which go-xdr uses to decide max size.

raymens:
@nullstyle What is the status of this issue?

nullstyle:
@raymens It's stalled.  None of us have looked at it since I originally implemented it.  I imagine I'll take another look at integrating it when I get around to reworking the xdr package or doing something else related to xdrgen.

lib/xdrgen/generators/go.rb
```diff
@@ -501,6 +523,25 @@ def switch_for(out, union, ident)
...
+        case field.declaration
+        when Xdrgen::AST::Declarations::String
+          size = field.declaration.resolved_size
+        when Xdrgen::AST::Declarations::Array
+          size = field.declaration.resolved_size unless field.declaration.fixed?
+        end
+
+        if size == "MAX_TX_PER_LEDGER"
+          binding.pry
+        end
```
bartekn: This `if` can now be removed.

PR #26, opened 2017-04-05
golang: Render size methods for opaque types

nullstyle:


bartekn:
Maybe I did something wrong but I can't see `XDRMaxSize` for `type Signature []byte` in a generated code. Also there is this file again: `.vscode/temp.sql`.

nullstyle:
You must have done something wrong...  I have a branch of the monorepo that has the right output.  I'm just waiting to merge this in so I can use the correct git hash in `Gemfile.lock`.

Also, thanks for the heads up on the sql method... I think maybe there's a bug in visual studio code, I swear they aren't in the commit preview.  I should probably just track down the extension that is creating it, too :)

bartekn:
I was building `master` 🤦‍♂️ . 

👍

PR #32, opened 2018-05-24
Add Scala generation support.

nbauernfeind:
It would greatly simplify stellar's scala sdk (see synesso/scala-stellar-sdk#13) to use an interface that more natively integrated with Scala than the Java API.

This change adds full generation support for scala. It was tested by generating source code for stellar-core's XDR classes and compiling them with scalac. More thorough testing is TBD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,12 +16,13 @@ Xdrgen is a very early project.  Aside from the test fixtures in
 are the .x files used for the
 [stellar-core project](https://github.com/stellar/stellar-core).
 
-Xdrgen presently supports three output languages:  ruby, javacript, java, and golang:
+Xdrgen presently supports the following output languages:
 
 - ruby: complete support
 - javascript: complete support
 - java: complete support
 - golang: currently using a fork of go-xdr, but has complete support
+- scala: complete support
 
 Testing is _very_ sparse, but will improve over time.
 
--- a/lib/xdrgen/generators.rb
+++ b/lib/xdrgen/generators.rb
@@ -6,6 +6,7 @@ module Xdrgen::Generators
   autoload :Go
   autoload :Javascript
   autoload :Java
+  autoload :Scala
 
   def self.for_language(language)
     const_get language.to_s.classify
--- a/lib/xdrgen/generators/scala.rb
+++ b/lib/xdrgen/generators/scala.rb
@@ -1,0 +1,475 @@
+module Xdrgen
+  module Generators
+
+    class Scala < Xdrgen::Generators::Base
+
+      def generate
+        render_lib
+        render_definitions(@top)
+      end
+
+      def render_lib
+        template = IO.read(__dir__ + "/scala/XdrDataInputStream.erb")
+        result = ERB.new(template).result binding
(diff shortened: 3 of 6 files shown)


Synesso:
Thanks Nate. This is looking good.

After running rspec I attempted to compile the results. There were 2 compilation errors & also the warning about classes overwriting on case-insensitive filesystems (I think we agreed this was not a dealbreaker, but I was under the impression that modern Windows systems are case-sensitive - however this doesn't seem to be the case. We probably should avoid case-insensitive class filename collisions).

```
$ for d in `ls`; do (echo $d; cd $d && scalac *.scala); done
block_comments.x
const.x
TestArray.scala:20: error: not found: value FOO
  def decode(stream: XdrDataInputStream): TestArray = TestArray((0 until FOO).map(_ => stream.readInt).toArray)
                                                                         ^
one error found
enum.x
nesting.x
MyUnion.scala:59: warning: Class MyXDR.MyUnion$One differs only in case from MyXDR.MyUnion$ONE. Such classes will overwrite one another on case-insensitive filesystems.
  case class One (
             ^
MyUnion.scala:67: warning: Class MyXDR.MyUnion$One$ differs only in case from MyXDR.MyUnion$ONE$. Such classes will overwrite one another on case-insensitive filesystems.
  object One {
         ^
MyUnion.scala:72: warning: Class MyXDR.MyUnion$Two differs only in case from MyXDR.MyUnion$TWO. Such classes will overwrite one another on case-insensitive filesystems.
  case class Two (
             ^
MyUnion.scala:82: warning: Class MyXDR.MyUnion$Two$ differs only in case from MyXDR.MyUnion$TWO$. Such classes will overwrite one another on case-insensitive filesystems.
  object Two {
         ^
four warnings found
optional.x
struct.x
test.x
OptHash1.scala:16: error: value encode is not a member of Option[MyXDR.Hash]
    self.encode(stream)
         ^
OptHash1.scala:20: error: type mismatch;
 found   : MyXDR.Hash
 required: Option[MyXDR.Hash]
  def decode(stream: XdrDataInputStream): OptHash1 = OptHash1(Hash.decode(stream))
                                                                         ^
OptHash2.scala:16: error: value encode is not a member of Option[MyXDR.Hash]
    self.encode(stream)
         ^
OptHash2.scala:20: error: type mismatch;
 found   : MyXDR.Hash
 required: Option[MyXDR.Hash]
  def decode(stream: XdrDataInputStream): OptHash2 = OptHash2(Hash.decode(stream))
                                                                         ^
four errors found
union.x
```

leighmcculloch:
@nbauernfeind  We're maintaining xdrgen generators outside of the xdrgen repo now, since we can't test them inside the repo very well or in the context of their SDK. There's discussion about this here:
- https://github.com/stellar/stellar-protocol/discussions/1738

All your existing code should work, it's just a matter of copying the ruby file containing the Scala generator into your own repo and add a couple Ruby files like Gemfile and calling the `Xdr::Compilation` type. There's an example of how to do this here:
- https://github.com/stellar/rs-stellar-xdr/pull/449

Issue #36, opened 2018-09-01
java generated code for current stellar xdr does not compile


jillesvangurp:
```
$ rake clean install && xdrgen -o tst -l java -n org.stellar.xdr ../stellar-core/src/xdr/*
xdrgen 0.0.1 built to pkg/xdrgen-0.0.1.gem.
xdrgen (0.0.1) installed.
jillesvangurp@jilless-MacBook-Pro ~/git/inbot/xdrgen[master]*
$ javac tst/*java
tst/AuthenticatedMessage.java:40: error: cannot find symbol
  stream.writeInt(encodedAuthenticatedMessage.getDiscriminant().getValue());
                                                               ^
  symbol:   method getValue()
  location: class Uint32
tst/AuthenticatedMessage.java:41: error: incompatible types: Uint32 cannot be converted to int
  switch (encodedAuthenticatedMessage.getDiscriminant()) {
         ^
tst/AuthenticatedMessage.java:51: error: incompatible types: Uint32 cannot be converted to int
  switch (decodedAuthenticatedMessage.getDiscriminant()) {
         ^
3 errors
```

PR #37, opened 2018-09-01
Java parent interface for xdr

jillesvangurp:
Add a common parent XdrElement interface.

Currently has a single convenience method so you can call encode directly on any XDR element without having to call the static method on the right class.

The main use for this is being able to reflect on classes and doing e.g. some kotlin extension functions.

Currently the java code does not compile because of #36 but this issue also occurs on master. My ruby is a bit rusty so would appreciate some help with that issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ mkmf.log
 spec/fixtures/hayashi/
 gen/
 .DS_Store
+.ruby-version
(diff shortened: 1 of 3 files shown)


jillesvangurp:
Merged changes from upstream and now everything compiles as it should.

@bartekn have a look at this one.

theaeolianmachine:
Merge this in @tamirms?

PR #38, opened 2018-09-12
Implement standard binary interfaces in Go output

bobg:
Adds `MarshalBinary` and `UnmarshalBinary` for each rendered definition, implementing the binary interfaces from [the standard encoding package](https://golang.org/pkg/encoding/).

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -177,7 +177,6 @@ module Xdrgen
           render_binary_interface out, name(defn)
         when AST::Definitions::Const ;
           render_const out, defn
-          render_binary_interface out, name(defn)
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/xdrgen/generators/go.rb
```diff
@@ -165,14 +165,19 @@ def render_definition(out, defn)
...
+          render_binary_interface out, name(defn)
         when AST::Definitions::Union ;
           render_union out, defn
+          render_binary_interface out, name(defn)
         when AST::Definitions::Typedef ;
           render_typedef out, defn
+          render_binary_interface out, name(defn)
         when AST::Definitions::Const ;
           render_const out, defn
+          render_binary_interface out, name(defn)
```
bartekn: This probably won't compile for `const`.

bobg: Good call, thanks. Fixed.

PR #41, opened 2018-09-24
Implement standard text encoding in Go output.

bobg:
Adds `MarshalText` and `UnmarshalText` for each rendered definition (using base64), implementing the standard `encoding.TextMarshaler` and `encoding.TextUnmarshaler` interfaces.

Among other things, this will allow generated XDR types to automatically marshal and unmarshal when embedded in JSON structures.

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -166,15 +166,19 @@ module Xdrgen
         when AST::Definitions::Struct ;
           render_struct out, defn
           render_binary_interface out, name(defn)
+          render_text_interface out, name(defn)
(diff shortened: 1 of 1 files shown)


bartekn:
I think it may be a breaking change for some data types. Ex. tests in https://github.com/stellar/go/pull/688 fail because `OperationTypeBumpSequence` is represented as a binary form (so `11` to XDR and then base64 encoded) instead of a simple numeric value when persisting the value in the database.

It looks like this can cause problems for enums and typedefs for basic Go numeric types (int, byte, float).

bobg:
Withdrawn.

PR #43, opened 2018-12-07
Scala support

Synesso:
This continues on from the good work started by @nbauernfeind and supercedes #32. At the time of opening the PR, the compilation errors against test fixtures have been resolved. Remaining work:

- [x] Resolve compilation warnings about case-insensitive classnames
- [x] Test validity of sources generated against https://github.com/stellar/stellar-core/tree/master/src/xdr


Synesso:
I've added union test case scenarios that arise when generating Stellar objects and were not already covered:

- Union nested unions
- Unions with void default arms
- Enum discriminated unions with incomplete arms and no default
- Type aliased primitive discriminated unions

Synesso:
This is ready for review.

PR #44, opened 2019-04-23
Support for standard type union discriminant with constant value

bartekn:
`Union.resolved_case` doesn't support unions with standard type discriminant and constant (`const`) values, ex:
```xdr
enum StellarValueType
{
    STELLAR_VALUE_BASIC = 0,
    STELLAR_VALUE_SIGNED = 1
};

union switch (int v)
{
case STELLAR_VALUE_BASIC:
    void;
    ...
```
To fix this, we are searching for identifier in namespace's enum constants if discriminant type has not been found.

EDIT, realized that it probably doesn't work for JS and Java too:
- [x] Update JS
- [x] Update Java

### DIFF
--- a/lib/xdrgen/ast/definitions/union.rb
+++ b/lib/xdrgen/ast/definitions/union.rb
@@ -19,8 +19,27 @@ module Xdrgen::AST
       end
 
       def resolved_case(kase)
-        found = discriminant_type.members.find{|m| m.name == kase.value_s}
-        raise "Case error:  #{kase} is not a member of #{discriminant_type.name}" if found.nil?
+        if discriminant_type.nil? then
+          # discriminant_type has not been found we need to search for the value in namespace's enum constants.
+          # It's a case where union discriminant is a standard type (like `int`):
+          #
+          # enum StellarValueType
+          # {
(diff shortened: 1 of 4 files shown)


nullstyle:
I'm cool with this generalized approach to solving the problem, but we should make sure to update all of the language generators and confirm they work before merging.

bartekn:
@nullstyle done. Unfortunately Java and JS solution is tricky.

In Java, the type of the value we use `switch` on is `Integer` and the types must match in `case`s. However, `StellarValueType.STELLAR_VALUE_BASIC` is not `Integer` and `StellarValueType.STELLAR_VALUE_BASIC.getValue()` is not constant (getting "Constant expression required" error).

In JS, I was getting "ChildUnion._switchOn.fromName is not a function" error because `_switchOn` on the union is equal `xdr.int()`.

In both cases I ended up with rendering actual numeric values of constants (`0` for `STELLAR_VALUE_BASIC` and `1` for `STELLAR_VALUE_SIGNED`). This is not ideal but is a working solution.

If you have a better idea how to solve this don't hesitate and change my code.

PR #46, opened 2019-06-12
Generate hashCode and equals methods for Java XDR classes

tamirms:


lib/xdrgen/generators/java.rb
```diff
@@ -23,24 +25,86 @@ def render_definitions(node)
         node.definitions.each(&method(:render_definition))
       end
 
+      def add_imports_for_definiton(defn, imports)
```
ire-and-curses: typo in function name

tamirms:
@bartekn does this pr look ok to you?

PR #47, opened 2019-11-06
Elixir generator

grossvogel:
I've been working with some colleagues on getting transaction support into the [Elixir Stellar client](https://github.com/revelrylabs/elixir-stellar-client), and we've published [this XDR lib](https://github.com/revelrylabs/exdr) as a part of that. `xdrgen` is the next step.

I haven't contributed to Stellar projects before (though I did just sign the agreement), so let me know what else I need iron out to get this incorporated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,8 @@ golang, and elixir:
 - javascript: complete support
 - java: complete support
 - golang: currently using a fork of go-xdr, but has complete support
-- elixir: support is experimental as the SDK is in early development
+- elixir: support is experimental as the SDK is in early development. Generated
+  code requires [:exdr](https://github.com/revelrylabs/exdr) in your deps
 
 Testing is _very_ sparse, but will improve over time.
 
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Build Status](https://travis-ci.org/stellar/xdrgen.svg)](https://travis-ci.org/stellar/xdrgen)
 [![Code Climate](https://codeclimate.com/github/stellar/xdrgen/badges/gpa.svg)](https://codeclimate.com/github/stellar/xdrgen)
 
-`xdrgen` is a code generator that takes XDR IDL files (`.x` files) as specfified
+`xdrgen` is a code generator that takes XDR IDL files (`.x` files) as specified
 in [RFC 4506](http://tools.ietf.org/html/rfc4506.html) and spits code out in
 various languages.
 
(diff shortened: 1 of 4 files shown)


ire-and-curses:
Hi @grossvogel! Thanks for this interesting PR. We're excited that you're working on an Elixir client (did you join the keybase channel #stellar.sdk-maintainers already?).

We aren't terribly familiar with Elixir. Would you add some tests for this code?

grossvogel:
@ire-and-curses Sure thing. I looked for generator-specific specs and failed to realize that the existing [generator spec](https://github.com/stellar/xdrgen/blob/master/spec/lib/xdrgen/generator_spec.rb) actually runs all of them. Now that I've seen that, I'll definitely add Elixir there and make sure it passes.

Beyond that, I'm having a hard time figuring out a good way to validate the output of the language-specific generator in the context of this project. In my [XDR lib](https://github.com/revelrylabs/exdr/blob/master/test/stellar/stellar_test.exs) I have a test that creates a transaction using the generated file and matches the binary representation against the same trx output from JS library, so that's the one that really gives me some confidence in this code. Anyway, I'd be happy to invest some time in additional tests here, especially if you have some hints on a good approach.

leighmcculloch:
We don't have tests like this in this repository but I think we should consider investing in tests like how @grossvogel described, where we exercise the generated code to build a common set of transactions and ensure that the final XDR message is exactly as expected and consistent across languages that we generate for. This isn't a small ask though. The first language we add these tests for will need to set the framework for how we can have multiple sets of tests and languages in this repository.

leighmcculloch:
I think it would be too big a task to add all those tests before merging this change and we can evaluate if we want to add those tests for all our languages.

Can we add the language to the `README.md` in this PR under the `Status` heading, and list it's status as `experimental` since it is new?

ire-and-curses:
Yes, this is a fair point. Also I feel bad putting a testing burden on you that isn't implemented for our other languages. How about this: let's add something to the README to indicate that Elixir support is "experimental". You should continue to write tests in your SDK, ideally covering all the Stellar operations + things like multi-sig. I think in practice this will be a good way to surface any bugs. In the medium term there are several ways this project might evolve and we can think more about tests here at that time.

ire-and-curses:
@abuiles can you do a quick sanity check on this? Please don't spend a ton of time, but you probably have the best combination of experience here. Then we can merge.

README.md
```diff
@@ -16,12 +16,14 @@ Xdrgen is a very early project.  Aside from the test fixtures in
...
 
-Xdrgen presently supports three output languages:  ruby, javacript, java, and golang:
+Xdrgen presently supports these output languages:  ruby, javacript, java,
+golang, and elixir:
 
 - ruby: complete support
 - javascript: complete support
 - java: complete support
 - golang: currently using a fork of go-xdr, but has complete support
+- elixir: support is experimental as the SDK is in early development
```
abuiles: @grossvogel let's mention here https://github.com/revelrylabs/exdr  and that it is required in your deps.

grossvogel: @abuiles Yes, I'll make that readme update for sure. Thank you for the reminder to implement `String.Chars` in the `exdr` library as well. It's something I've meant to do but kept forgetting. I created [an issue](https://github.com/revelrylabs/exdr/issues/17) and hope to get to it in the next couple of weeks.

abuiles: @grossvogel awesome,  I think we can merge this once we have the link to `exdr`

grossvogel: OK, I added the note. Hopefully the wording is OK 😸

abuiles: awesome, thanks!

PR #48, opened 2019-12-10
Represent XDR strings as bytes

tamirms:
According to https://github.com/stellar/go/issues/2022 XDR strings are equivalent to XDR opaque fields. The Java XDR code generator represents XDR strings as Java Strings. This representation is incorrect because an XDR string has no restrictions on the byte format. It is possible for an XDR string to contain invalid ASCII or invalid unicode bytes. The safest representation of an XDR string is a byte array.

### DIFF
--- a/lib/xdrgen/generators/java.rb
+++ b/lib/xdrgen/generators/java.rb
@@ -683,19 +683,19 @@ module Xdrgen
           out.puts <<-EOS.strip_heredoc
             #{value}.#{member.name} = new #{type_string member.type}[#{member.name}size];
             for (int i = 0; i < #{member.name}size; i++) {
-              #{value}.#{member.name}[i] = #{decode_type member.declaration.type};
+              #{value}.#{member.name}[i] = #{decode_type member.declaration};
             }
           EOS
         else
-          out.puts "#{value}.#{member.name} = #{decode_type member.declaration.type};"
+          out.puts "#{value}.#{member.name} = #{decode_type member.declaration};"
         end
         if member.type.sub_type == :optional
           out.puts "}"
         end
       end
 
-      def decode_type(type)
-        case type
+      def decode_type(decl)
+        case decl.type
         when AST::Typespecs::Int ;
           "stream.readInt()"
         when AST::Typespecs::UnsignedInt ;
@@ -713,13 +713,13 @@ module Xdrgen
         when AST::Typespecs::Bool ;
           "stream.readInt() == 1 ? true : false"
         when AST::Typespecs::String ;
-          "XdrString.decode(stream)"
+          "XdrString.decode(stream, #{decl.size})"
         when AST::Typespecs::Simple ;
-          "#{name type.resolved_type}.decode(stream)"
+          "#{name decl.type.resolved_type}.decode(stream)"
         when AST::Concerns::NestedDefinition ;
-          "#{name type}.decode(stream)"
+          "#{name decl.type}.decode(stream)"
         else
-          raise "Unknown typespec: #{type.class.name}"
+          raise "Unknown typespec: #{decl.type.class.name}"
         end
       end
 
--- a/lib/xdrgen/generators/java.rb
+++ b/lib/xdrgen/generators/java.rb
@@ -22,6 +22,10 @@ module Xdrgen
         template = IO.read(__dir__ + "/java/XdrElement.erb")
         result = ERB.new(template).result binding
         @output.write  "XdrElement.java", result
+
+        template = IO.read(__dir__ + "/java/XdrString.erb")
+        result = ERB.new(template).result binding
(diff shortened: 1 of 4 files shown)


lib/xdrgen/generators/java.rb
```diff
@@ -637,8 +643,6 @@ def encode_type(type, value)
           raise "cannot render quadruple in golang"
         when AST::Typespecs::Bool ;
           "stream.writeInt(#{value} ? 1 : 0)"
-        when AST::Typespecs::String ;
-          "stream.writeString(#{value})"
```
leighmcculloch: With this change the `writeString` and `readString` functions become unused I think. Should we continue to use them and redefine inside those functions how we write and read the string? Or should we remove those functions?

https://github.com/stellar/xdrgen/blob/c683684abb3cb2c8b8c3dc12dc7c6cc9bf1e3b60/lib/xdrgen/generators/java/XdrDataOutputStream.erb#L17-L21

https://github.com/stellar/xdrgen/blob/c683684abb3cb2c8b8c3dc12dc7c6cc9bf1e3b60/lib/xdrgen/generators/java/XdrDataInputStream.erb#L24-L29

tomerweller:
@tamirms can you attach an example of how a generated String class will look like? It's hard to understand from the generator. 

I think we should add a `toString()` method to these classes that makes a best effort to stringify the byte-array. If it's not utf8 it can throw an exception or maybe even return a hex representation.

lib/xdrgen/generators/java.rb
tamirms: I pushed a commit which removes the methods

tamirms:
@tomerweller 

checkout https://github.com/stellar/java-stellar-sdk/pull/259 to see what the generated output looks like. The Java sdk converts the bytes to a string. If the bytes are not valid utf 8 the Java String class will convert the bytes to valid utf8.

tomerweller:
@tamirms my point was that every time you set a String or read one you will need to encode/decode utf8, so it might save some headache for consumers to have that functionality in the generated string class rather that in the sdk level. 

So for example (and this is a matter of java style), having another ctor that takes in a `String` and a `toString()` method that handle byte array utf8 conversions behind the scenes, could save consumers the trouble of thinking in byte arrays and utf8. 

Some references from the mentioned pr:
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-420db4b3433ba4cb9daba85f6dd6a5a8R46
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-420db4b3433ba4cb9daba85f6dd6a5a8R73
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-11149bfd1ec2bd50ce92029356b7f057R18
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-11149bfd1ec2bd50ce92029356b7f057R29
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-fca6cbe50b2719ae470794aa856f6660R153
https://github.com/stellar/java-stellar-sdk/pull/259/files#diff-fca6cbe50b2719ae470794aa856f6660R211

tamirms:
@tomerweller please check the most recent commit

lib/xdrgen/generators/java/XdrString.erb
```diff
@@ -0,0 +1,57 @@
+package <%= @namespace %>;
+
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.Arrays;
+
+public class XdrString implements XdrElement {
+    private byte[] bytes;
+    private String text;
```
tomerweller: I'm not a fan of keeping the same state in two variables. I'd get rid of the `text` member and just keep the `bytes`, as that's the closest representation to what's on the wire.

lib/xdrgen/generators/java/XdrString.erb
```diff
@@ -0,0 +1,57 @@
...
+          return false;
+        }
+
+        XdrString other = (XdrString) object;
+        return Arrays.equals(this.bytes, other.bytes);
+    }
+
+    @Override
+    public String toString() {
+        return this.text;
```
tomerweller: Following on the previous comment, you can construct a new String here from `bytes`, instead of keeping `text` around.

lib/xdrgen/generators/java/XdrString.erb
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public XdrString(String text) {
+        this.bytes = text.getBytes(Charset.forName("UTF-8"));
+    }
+
+    @Override
+    public void encode(XdrDataOutputStream stream) throws IOException {
+        stream.writeInt(this.bytes.length);
+        stream.write(this.bytes, 0, this.bytes.length);
```
bartekn: We should probably add a `maxSize` check here too.

PR #52, opened 2020-09-21
Downcase value when comparing if the source is plural or singular.

abuiles:
While generating protocol 14 XDRs,  I noticed a weird output for CAP33 related enums and ops.

The following:

```
 beginSponsoringFutureReserf: 16,
 endSponsoringFutureReserf: 17,
```

Should have been 

```
 beginSponsoringFutureReserves: 16,
 endSponsoringFutureReserves: 17,
```

After investigating, I found a bug in this library, which was comparing the pluralized version of the name with the original, the problem is that it didn't keep into account the letter casing - leading to incorrect results in scenarios like this one

```
"BEGIN_SPONSORING_FUTURE_RESERVEs" == "BEGIN_SPONSORING_FUTURE_RESERVES"
```

This PR fixes this issue, making both values down case.

### DIFF
--- a/lib/xdrgen/generators/elixir.rb
+++ b/lib/xdrgen/generators/elixir.rb
@@ -164,7 +164,7 @@ module Xdrgen
         parent = name named.parent_defn if named.is_a?(AST::Concerns::NestedDefinition)
 
         # NOTE: classify will strip plurality, so we restore it if necessary
-        plural = named.name.pluralize == named.name
+        plural = named.name.downcase.pluralize == named.name.downcase
         base   = named.name.underscore.classify
         result = plural ? base.pluralize : base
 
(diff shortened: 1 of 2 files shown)


lib/xdrgen/generators/elixir.rb
```diff
@@ -164,18 +164,15 @@ def name(named)
...
         base   = named.name.underscore.classify
         result = plural ? base.pluralize : base
 
         "#{parent}#{result}"
       end
 
       def const_name(named)
-        # NOTE: classify will strip plurality, so we restore it if necessary
-        plural = named.name.pluralize == named.name
-        base   = named.name.underscore.upcase
```
abuiles, line 177: we  don't pluralize base, so this check is actually unnecessary.

2opremio (approved):

PR #53, opened 2020-10-06
Fix grammar to support multiline union header

### LINKED ISSUE #50: Handle unions with multiple lines
It seems like we can't properly parse unions when the discriminant is specified in a different line.  The following line yields an error:

```
union PathPaymentStrictReceiveResult switch (
    PathPaymentStrictReceiveResultCode code)
{
case PATH_PAYMENT_STRICT_RECEIVE_SUCCESS:
    struct
    {
        ClaimOfferAtom offers<>;
        SimplePaymentResult last;
    } success;
case PATH_PAYMENT_STRICT_RECEIVE_NO_ISSUER:
    Asset noIssuer; // the asset that caused the error
default:
    void;
};
```

nebolsin:
Fixes #50.

Otherwise generate is failing for cases like this: https://github.com/stellar/stellar-core/blob/master/src/xdr/Stellar-transaction.x#L703-L704

```
union PathPaymentStrictReceiveResult switch (
    PathPaymentStrictReceiveResultCode code)
{
case PATH_PAYMENT_STRICT_RECEIVE_SUCCESS:
    struct
    {
        ClaimOfferAtom offers<>;
        SimplePaymentResult last;
    } success;
case PATH_PAYMENT_STRICT_RECEIVE_NO_ISSUER:
    Asset noIssuer; // the asset that caused the error
default:
    void;
};
```

### DIFF
--- a/lib/xdrgen/grammar/union.treetop
+++ b/lib/xdrgen/grammar/union.treetop
@@ -12,7 +12,11 @@ grammar XdrUnionGrammar
   rule union_body
     "switch"
     space?
-    "(" discriminant:declaration ")"
+    "("
+    space?
+    discriminant:declaration
+    space?
+    ")"
     space?
     "{"
     space?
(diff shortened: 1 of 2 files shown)


abuiles:
Thanks!

PR #60, opened 2021-01-27
Fix C# unsigned int generation

fracek:
This fixes a problem in the `Encode` and `Decode` methods when generating unsigned types (e.g. `UInt32.cs`).

Now the code generated should match the one living in `dotnet-stellar-sdk` line by line.

### DIFF
--- a/lib/xdrgen/generators/csharp.rb
+++ b/lib/xdrgen/generators/csharp.rb
@@ -386,11 +386,11 @@ module Xdrgen
         when AST::Typespecs::Int
           "stream.WriteInt(#{value})"
         when AST::Typespecs::UnsignedInt
-          "stream.WriteInt(#{value})"
+          "stream.WriteUInt(#{value})"
         when AST::Typespecs::Hyper
           "stream.WriteLong(#{value})"
         when AST::Typespecs::UnsignedHyper
-          "stream.WriteLong(#{value})"
+          "stream.WriteULong(#{value})"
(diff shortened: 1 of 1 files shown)


lib/xdrgen/generators/csharp.rb
```diff
@@ -386,11 +386,11 @@ def encode_type(type, value)
         when AST::Typespecs::Int
           "stream.WriteInt(#{value})"
         when AST::Typespecs::UnsignedInt
-          "stream.WriteInt(#{value})"
+          "stream.WriteUInt(#{value})"
```
leighmcculloch, line 389: Where is the `WriteUint` function defined? I can't find it in:

https://github.com/stellar/xdrgen/blob/9dacf24a73b3e73105e9272ccf333704dcc82333/lib/xdrgen/generators/java/XdrDataOutputStream.erb#L8-L9

Or in the docs for `java.io.DataOutputStream`:

https://docs.oracle.com/javase/7/docs/api/java/io/DataOutputStream.html

I'm very rusty on C# though, so I'm sure I'm missing something.

leighmcculloch: It may be beneficial to build that logic into the generator so it is possible to use it to generate other XDR formats, but that is unrelated to this change and this PR.

leighmcculloch:
Thanks @fracek @elucidsoft!

PR #63, opened 2021-08-19
Update CI provider from Travis to Github Actions

marcelosalloum:
Update from travis CI to Github Actions

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


.github/workflows/ruby.yml
```diff
@@ -0,0 +1,24 @@
+name: Ruby
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
leighmcculloch, line 7: Let's run builds on all PRs, regardless of their base branch.
```suggestion
on:
  push:
    branches: [ master ]
  pull_request:
```

leighmcculloch (approved):
One suggested change, otherwise looks good. Good idea to add v3 of ruby too.

.github/workflows/ruby.yml
marcelosalloum: Applied in https://github.com/stellar/xdrgen/pull/63/commits/2848f73ff644349cb6540796f1c801f9ff744229.

PR #67, opened 2021-11-10
Generate MarhsalXDR() instead of EncodeTo()

2opremio:
xdrgen counterpart to https://github.com/stellar/go-xdr/pull/17

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -18,13 +18,13 @@ module Xdrgen
       def render_typedef(out, typedef)
         # Typedefs that wrap a pointer type are not well supported in Go because
         # Go does not allow pointer types to have methods. This prevents us from
-        # defining the EncodeTo method on these types which is very inconvenient
+        # defining the MarshalXDR method on these types which is very inconvenient
         # for the render functions that generate structs that contain these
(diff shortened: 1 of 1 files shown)


lib/xdrgen/generators/go.rb
```diff
@@ -336,73 +332,74 @@ def render_union(out, union)
...
       end
 
-      def render_struct_encode_to_interface(out, struct)
+      def render_struct_marshal_xdr_interface(out, struct)
         name = name(struct)
-        out.puts "// EncodeTo encodes this value using the Encoder."
-        out.puts "func (s *#{name}) EncodeTo(e *xdr.Encoder) error {"
-        out.puts "  var err error"
+        out.puts "// MarshalXDR implements xdr.Marshaler."
+        out.puts "func (s #{name}) MarshalXDR(e *xdr.Encoder) (int, error) {"
```
2opremio, line 338: Unfortunately, I am forced to call by value (note the lack of `*`) when switching to `MarshallXDR)()` since, otherwise, `xdr.Encoder()` cannot invoke the  method on unadresable values.

This implies calling by value, which can (and as I observed, it does) cause allocations, betraying the purpose of doing all this (which is increasing performance).

2opremio:
I'm closing this PR, due to the reasons outlined above.

PR #71, opened 2021-11-11
Generate `DecodeFrom()` methods for Golang

2opremio:
This is the decoding counterpart of https://github.com/stellar/xdrgen/pull/65

There is a dramatic performance improvement: ~13x speedup and and ~11x  allocation reduction.

Check https://github.com/stellar/go/pull/4064 for details on the speedup and the generated code.

TODO:
- [ ] Make `EncodeTo()`-related code consistent with the changes in this PR. I intentionally didn't touch `EncodeTo`-code here in order to avoid noise (otherwise it will be more difficult to read the generated code at https://github.com/stellar/go/pull/4064 ). I will update them in a separate PR.

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -566,7 +566,9 @@ module Xdrgen
           out2.puts "  return n, nil"
           out2.string
         end
-        out.puts "  return n, fmt.Errorf(\"#{name(union.discriminant)} (#{reference union.discriminant.type}) switch value '%d' is not valid for union #{name}\", u.#{name(union.discriminant)})"
+        unless union.default_arm.present?
+          out.puts "  return n, fmt.Errorf(\"#{name(union.discriminant)} (#{reference union.discriminant.type}) switch value '%d' is not valid for union #{name}\", u.#{name(union.discriminant)})"
+        end
         out.puts "}"
         out.break
       end
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -532,9 +532,10 @@ module Xdrgen
         out.puts "// DecodeFrom decodes this value using the Decoder."
         out.puts "func (s *#{name}) DecodeFrom(d *xdr.Decoder) error {"
         out.puts "  var err error"
+        declared_variables = []
         struct.members.each do |m|
           mn = name(m)
-          render_decode_from_body(out, "s.#{mn}", m.type, self_encode: false)
+          render_decode_from_body(out, "s.#{mn}", m.type, declared_variables: declared_variables, self_encode: false)
         end
         out.puts "  return nil"
         out.puts "}"
@@ -544,19 +545,21 @@ module Xdrgen
       def render_union_decode_from_interface(out, union)
         name = name(union)
         out.puts "// DecodeFrom decodes this value using the Decoder."
-        out.puts "func (s *#{name}) DecodeFrom(d *xdr.Decoder) error {"
-        out.puts "  discriminant, _, err := d.DecodeInt()"
+        out.puts "func (u *#{name}) DecodeFrom(d *xdr.Decoder) error {"
+        out.puts "  disc, _, err := d.DecodeInt()"
         out.puts "  if err != nil {"
         out.puts "    return err"
         out.puts "  }"
-        out.puts "  s.#{name(union.discriminant)} = #{reference union.discriminant.type}(discriminant)"
-        switch_for(out, union, "s.#{name(union.discriminant)}") do |arm, kase|
+        out.puts "  u.#{name(union.discriminant)} = #{reference union.discriminant.type}(disc)"
+        switch_for(out, union, "u.#{name(union.discriminant)}") do |arm, kase|
           out2 = StringIO.new
           if arm.void?
             "// Void"
           else
             mn = name(arm)
-            render_decode_from_body(out2, "(*s.#{mn})", arm.type, self_encode: false)
+            type = arm.type
+            out2.puts "u.#{mn} = new(#{reference arm.type})"
+            render_decode_from_body(out2, "(*u.#{mn})",type, declared_variables: [], self_encode: false)
             out2.string
           end
         end
@@ -569,11 +572,11 @@ module Xdrgen
         name = name(typedef)
         type = typedef
         out.puts "// DecodeFrom decodes this value using the Decoder."
-        out.puts "func (s *#{name}) DecodeFrom(d *xdr.Decoder) error {"
+        out.puts "func (e *#{name}) DecodeFrom(d *xdr.Decoder) error {"
         out.puts "  var err error"
-        out.puts "  var e int32"
-        render_decode_from_body(out, "e", type, self_encode: true)
-        out.puts "  *s = #{name}(e)"
+        out.puts "  var i int32"
+        render_decode_from_body(out, "i", type, declared_variables: [], self_encode: true)
+        out.puts "  *e = #{name}(i)"
         out.puts "  return nil"
         out.puts "}"
         out.break
@@ -585,10 +588,40 @@ module Xdrgen
         out.puts "// DecodeFrom decodes this value using the Decoder."
         out.puts "func (s *#{name}) DecodeFrom(d *xdr.Decoder) error {"
         out.puts "  var err error"
-        render_decode_from_body(out, "s", type, self_encode: true)
+        var = "s"
+        sub_var_type = ""
+        # TODO: maybe use a reference() call
+        case type
+        when AST::Typespecs::UnsignedHyper
+          sub_var_type = "uint64"
+        when AST::Typespecs::Hyper
+          sub_var_type = "int64"
+        when AST::Typespecs::UnsignedInt
+          sub_var_type = "uint32"
+        when AST::Typespecs::Int, AST::Definitions::Enum
+          sub_var_type = "int32"
+        when AST::Typespecs::String
+          sub_var_type = "string"
+        end
+        if (type.is_a?(AST::Typespecs::Opaque) && !type.fixed?) || (type.is_a?(AST::Typespecs::Simple) && type.sub_type == :var_array)
+            var = "(*s)"
+        end
+        if !sub_var_type.empty?
+          out.puts "  var v #{sub_var_type}"
+          var = "v"
+        end
+        render_decode_from_body(out, var, type, declared_variables: [], self_encode: true)
+        out.puts "  *s = #{name}(v)" if !sub_var_type.empty?
         out.puts "  return nil"
         out.puts "}"
         out.break
+      end
+
+      def render_variable_declaration(out, indent, var, type, declared_variables:)
+        if !declared_variables.include?var
+           out.puts "#{indent}var #{var} #{type}"
+           declared_variables.append(var)
+        end
       end
 
       # render_decode_from_body assumes there is an `d` variable containing an
@@ -596,20 +629,22 @@ module Xdrgen
       # encode.
       def render_decode_from_body(out, var, type, declared_variables:, self_encode:)
         optional = type.sub_type == :optional
+        close_optional_within = false
         if optional
+          render_variable_declaration(out, "  ", 'b', "bool", declared_variables: declared_variables)
           out.puts "  b, _, err = d.DecodeBool()"
           out.puts "  if err != nil {"
           out.puts "    return err"
           out.puts "  }"
           out.puts "  #{var} = nil"
           out.puts "  if b {"
-          var = "(*#{var})"
+          out.puts "     #{var} = new(#{name type})"
         end
         case type
         when AST::Typespecs::UnsignedHyper
           out.puts "  #{var}, _, err = d.DecodeUhyper()"
         when AST::Typespecs::Hyper
-          out.puts "  #{var}, _, err = d.DecoderHyper()"
+          out.puts "  #{var}, _, err = d.DecodeHyper()"
         when AST::Typespecs::UnsignedInt
           out.puts "  #{var}, _, err = d.DecodeUint()"
         when AST::Typespecs::Int, AST::Definitions::Enum
@@ -621,7 +656,7 @@ module Xdrgen
           if type.fixed?
             out.puts "  _, err = d.DecodeFixedOpaqueInplace(#{var}[:])"
           else
-            # TODO: use maxsize annotation
+            # TODO: check maxsize annotation abd use that instead of 0
             out.puts "  #{var}, _, err = d.DecodeOpaque(0)"
           end
         when AST::Typespecs::Simple
@@ -629,30 +664,30 @@ module Xdrgen
           when :simple, :optional
             optional_within = type.is_a?(AST::Identifier) && type.resolved_type.sub_type == :optional
             if optional_within
+              render_variable_declaration(out, "  ", 'b', "bool", declared_variables: declared_variables)
               out.puts "  b, _, err = d.DecodeBool()"
               out.puts "  if err != nil {"
               out.puts "    return err"
               out.puts "  }"
               out.puts "  #{var} = nil"
               out.puts "  if b {"
-              var = "(*#{var})"
+              out.puts "     #{var} = new(#{name type.resolved_type.declaration.type})"
             end
-            var = "#{name type}(#{var})" if self_encode
+            var = "(*#{name type})(#{var})" if self_encode
             out.puts "  err = #{var}.DecodeFrom(d)"
-            if optional_within
-              out.puts "  }"
-            end
+            close_optional_within = optional_within
           when :array
             out.puts "  for i := 0; i < len(#{var}); i++ {"
             element_var = "#{var}[i]"
             optional_within = type.is_a?(AST::Identifier) && type.resolved_type.sub_type == :optional
             if optional_within
-              out.puts "    b, _, err = d.DecodeBool()"
+              out.puts "    var eb bool"
+              out.puts "    eb, _, err = d.DecodeBool()"
               out.puts "    if err != nil {"
               out.puts "      return err"
               out.puts "    }"
               out.puts "    #{var} = nil"
-              out.puts "    if b {"
+              out.puts "    if eb {"
               var = "(*#{element_var})"
             end
             out.puts "      err = #{element_var}.DecodeFrom(d)"
@@ -664,6 +699,7 @@ module Xdrgen
             end
             out.puts "  }"
           when :var_array
+            render_variable_declaration(out, "  ", 'l', "uint32", declared_variables: declared_variables)
             out.puts "  l, _, err = d.DecodeUint()"
             out.puts "  if err != nil {"
             out.puts "    return err"
@@ -673,12 +709,14 @@ module Xdrgen
             element_var = "#{var}[i]"
             optional_within = type.is_a?(AST::Identifier) && type.resolved_type.sub_type == :optional
             if optional_within
-              out.puts "    b, _,  err = d.DecodeBool()"
+              out.puts "    var eb bool"
+              out.puts "    eb, _,  err = d.DecodeBool()"
               out.puts "    if err != nil {"
               out.puts "      return err"
               out.puts "    }"
               out.puts "    #{element_var} = nil"
-              out.puts "    if b {"
+              out.puts "    if eb {"
+              out.puts "       #{element_var} = new(#{name type.resolved_type.declaration.type})"
               var = "(*#{element_var})"
             end
             out.puts "      err = #{element_var}.DecodeFrom(d)"
@@ -701,12 +739,16 @@ module Xdrgen
         else
           out.puts "  _, err = e.Encode(#{var})"
         end
-        if optional
-          out.puts "  }"
-        end
         out.puts "  if err != nil {"
         out.puts "    return err"
         out.puts "  }"
+        if close_optional_within
+           # otherwise the error checking above will be rendered outside the optional_within's if block
+           out.puts "  }"
+        end
+        if optional
+           out.puts "  }"
+        end
       end
 
       def render_binary_interface(out, name)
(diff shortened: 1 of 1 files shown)


2opremio:
It seems there are still a few minor problems ([decoding an empty value instead of nil in some cases](https://app.circleci.com/pipelines/github/stellar/go/9830/workflows/3e6c5739-76c4-4000-a756-bb620934ee28/jobs/57532)). I will look into that soon

2opremio:
OK, that's fixed and all tests are passing.

2opremio:
@bartekn could you take a quick look at the failure above:

```
ERRO[2021-11-12T00:24:58.012Z] Error in ingestion state machine              current_state="verifyRange(fromLedger=10000063, toLedger=10000127, verifyState=true)" error="Error preparing range: Error fast-forwarding to 10000063: stellar core exited unexpectedly" next_state=stop pid=430 service=ingest
INFO[2021-11-12T00:24:58.012Z] Shut down                                     pid=430 service=ingest
Error preparing range: Error fast-forwarding to 10000063: stellar core exited unexpectedly
```

I don't understand why stellar core fails and there is no context.

2opremio:
OK, I think I know what's happening. I shouldn't be using `ReadAll()`

There was a bug in the captive core backend hiding the error https://github.com/stellar/go/pull/4066

2opremio:
OK, the verify-range problem is fixed. I am going to run verify-range in  the last 100k ledgers 🤞

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,256 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+
+      def render_union_decode_from_interface(out, union)
+        name = name(union)
+        out.puts "// DecodeFrom decodes this value using the Decoder."
+        out.puts "func (u *#{name}) DecodeFrom(d *xdr.Decoder) (int, error) {"
+        out.puts "  disc, n, err := d.DecodeInt()"
+        out.puts "  if err != nil {"
+        out.puts "    return 0, err"
+        out.puts "  }"
+        out.puts "  u.#{name(union.discriminant)} = #{reference union.discriminant.type}(disc)"
```
bartekn, line 558: I mentioned this on Slack but you should check here if `disc` is actually a valid value for this union. Something like what `go-xdr` is doing here: https://github.com/stellar/go-xdr/blob/8017fc4bdfee412460b17f99d7d56fde3ab29047/xdr3/decode.go#L541-L555 Without it malicious actor could pass invalid value crashing a decoder.

2opremio: We check for it, see the generated code.

2opremio:
Note for reviewers, please check the code generated at stellar/go#4064 before reviewing the code here.

In fact, it's vital to review the code generated before reviewing this PR itself.

lib/xdrgen/generators/go.rb
2opremio: This does the right thing according to the spec. The type has a default void value, which means that the type is considered void if the discriminant is different to `CreateAccountResultCodeCreateAccountSuccess`. You can check the XDR spec.

2opremio: The definition of the type is:

```
// CreateAccountResult is an XDR Union defines as:
//
//   union CreateAccountResult switch (CreateAccountResultCode code)
//    {
//    case CREATE_ACCOUNT_SUCCESS:
//        void;
//    default:
//        void;
//    };
//
```

So, as unintuitive as this may seem, you can use whichever value we want other than `CREATE_ACCOUNT_SUCCESS` and it will still be considered valid.

2opremio: Note that for unions without the default arm, we also do the right thing thanks to the error returned at the end.

2opremio: I think that using default arms is questionable. We can chose to change the XDR definitions to disable default arms, but that's a totally different issue.

bartekn: I disagree this is a valid code according to xdr specification. See https://datatracker.ietf.org/doc/html/rfc4506.html#section-4.15:
```
Discriminated unions
   are declared as follows:

         union switch (discriminant-declaration) {
         case discriminant-value-A:
            arm-declaration-A;
         case discriminant-value-B:
            arm-declaration-B;
         ...
         default: default-declaration;
         } identifier;

   Each "case" keyword is followed by a legal value of the discriminant.
   The default arm is optional. If it is not specified, then a valid
   encoding of the union cannot take on unspecified discriminant values.

```
Each case needs to be followed by a _legal_ value of the discriminant and it cannot take on unspecified discriminant values.

2opremio: Note how `ArmForSwitch()` (which is what `go-xdr` uses for union arm checking) does the exact same thing:

```go
func (u CreateAccountResult) ArmForSwitch(sw int32) (string, bool) {
	switch CreateAccountResultCode(sw) {
	case CreateAccountResultCodeCreateAccountSuccess:
		return "", true
	default:
		return "", true
	}
}
```

bartekn: `ArmForSwitch` is not used in the code I linked here: https://github.com/stellar/go-xdr/blob/8017fc4bdfee412460b17f99d7d56fde3ab29047/xdr3/decode.go#L541-L555. I think spec is pretty clear on the matter we discuss here and the old code returns error on invalid discriminant values.

2opremio: Here is another example without default value:

```
// Memo is an XDR Union defines as:
//
//   union Memo switch (MemoType type)
//    {
//    case MEMO_NONE:
//        void;
//    case MEMO_TEXT:
//        string text<28>;
//    case MEMO_ID:
//        uint64 id;
//    case MEMO_HASH:
//        Hash hash; // the hash of what to pull from the content server
//    case MEMO_RETURN:
//        Hash retHash; // the hash of the tx you are rejecting
//    };
//
```

Note how both `ArmForSwitch` do return an error in this case:

```go
// ArmForSwitch returns which field name should be used for storing
// the value for an instance of Memo
func (u Memo) ArmForSwitch(sw int32) (string, bool) {
	switch MemoType(sw) {
	case MemoTypeMemoNone:
		return "", true
	case MemoTypeMemoText:
		return "Text", true
	case MemoTypeMemoId:
		return "Id", true
	case MemoTypeMemoHash:
		return "Hash", true
	case MemoTypeMemoReturn:
		return "RetHash", true
	}
	return "-", false
}
```

`go-xdr` will interpret "-" as an error at https://github.com/stellar/go-xdr/blob/master/xdr3/decode.go#L577-L586 

```
// DecodeFrom decodes this value using the Decoder.
func (u *Memo) DecodeFrom(d *xdr.Decoder) (int, error) {
	disc, n, err := d.DecodeInt()
	if err != nil {
		return 0, err
	}
	u.Type = MemoType(disc)
	switch MemoType(u.Type) {
	case MemoTypeMemoNone:
		// Void
		return n, nil
	case MemoTypeMemoText:
		u.Text = new(string)
		var nTmp int
		(*u.Text), nTmp, err = d.DecodeString(28)
		n += nTmp
		if err != nil {
			return n, err
		}
		return n, nil
	case MemoTypeMemoId:
		u.Id = new(Uint64)
		var nTmp int
		nTmp, err = (*u.Id).DecodeFrom(d)
		n += nTmp
		if err != nil {
			return n, err
		}
		return n, nil
	case MemoTypeMemoHash:
		u.Hash = new(Hash)
		var nTmp int
		nTmp, err = (*u.Hash).DecodeFrom(d)
		n += nTmp
		if err != nil {
			return n, err
		}
		return n, nil
	case MemoTypeMemoReturn:
		u.RetHash = new(Hash)
		var nTmp int
		nTmp, err = (*u.RetHash).DecodeFrom(d)
		n += nTmp
		if err != nil {
			return n, err
		}
		return n, nil
	}
	return n, fmt.Errorf("Type (MemoType) switch value '%d' is not valid for union Memo", u.Type)
}
```

bartekn: A few lines above the code you linked (that I linked directly in 2 comments above) there's a check using `ValidEnum` method that ensures that discriminant value is valid and I think we need the same thing in the new code. This is required by XDR spec and without this check invalid XDR can be decoded without errors which should not happen.

2opremio: @bartekn I just updated the code to remove redundant return values when there is a default branch (In case you notice the code changed)

bartekn: OK, I read entire thread again. It seems that the old decoder's code is wrong then (because it doesn't allow any value on `default` case). Sorry for the noise!

2opremio: Ah, good point! I will incorporate that to the code.

I also noticed the `DecodeEnum` function which validates enums, I will incorporate that too.

2opremio: OK, this should be fixed, PTAL

2opremio:
verify range succeeded! 🥳 

```
Status of auto-second-half-job28279:
{
  "STARTING": 0,
  "FAILED": 0,
  "RUNNING": 0,
  "SUCCEEDED": 9,
  "RUNNABLE": 0,
  "SUBMITTED": 0,
  "PENDING": 0
}
```

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+        var = "s"
+        sub_var_type = ""
+        case type
+        when AST::Typespecs::UnsignedHyper, AST::Typespecs::Hyper, AST::Typespecs::UnsignedInt, AST::Typespecs::Int, AST::Typespecs::String
+          sub_var_type = reference(type)
+        end
+        if (type.is_a?(AST::Typespecs::Opaque) && !type.fixed?) || (type.is_a?(AST::Typespecs::Simple) && type.sub_type == :var_array)
+            var = "(*s)"
+        end
+        if !sub_var_type.empty?
```
paulbellamy, line 612: nit: more idiomatic ruby way for `if !foo` is `unless foo`...
```suggestion
        unless sub_var_type.empty?
```

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+            out.puts "  nTmp, err = #{var}.DecodeFrom(d)"
+            out.puts tail
+            if optional_within
+              out.puts "  }"
+            end
+          when :array
+            out.puts "  for i := 0; i < len(#{var}); i++ {"
+            element_var = "#{var}[i]"
+            optional_within = type.is_a?(AST::Identifier) && type.resolved_type.sub_type == :optional
+            if optional_within
```
paulbellamy, line 699: Intuitively, it seems like this optional block (and above) should be able to be handled with a recursive data-type, like a `Option<Foo>` type-thing, and having a `render_optional` method... 🤔 but this is probably good enough for now.

paulbellamy:
also, wdyt of, instead of doing `puts "     some spaces for indentation"`, just doing `puts "no indentation spaces"`, and running `gofmt` on it at the end?

lib/xdrgen/generators/go.rb
paulbellamy: I think I'm thinking something like:
```ruby

def render_decode_from_body(out, var, type, declared_variables:, self_encode:)
  # ...
  if type.sub_type == :optional
    return render_optional(out, var, type, type, declared_variables: declared_variables, self_encode: self_encode) do |inner_type|
      render_decode_from_body(out, var, inner_type, declared_variables: declared_variables, self_encode: self_encode)
    end
  end
  # ...
end

def render_optional(out, var, type, declared_variables:, self_encode:)
  render_variable_declaration(out, "  ", 'b', "bool", declared_variables: declared_variables)
  out.puts "  b, nTmp, err = d.DecodeBool()"
  out.puts tail
  out.puts "  #{var} = nil"
  out.puts "  if b {"
  out.puts "     #{var} = new(#{name type.resolved_type.declaration.type})"
  # TODO: Unwrap the inner_type (without the optional wrapper) here
  yield inner_type
  out.puts "  }"
end
```

But, actually you could do a similar block thing for arrays as well, I think...

2opremio:
> also, unrelated to this PR, but wdyt of, instead of doing `puts " some spaces for indentation"`, just doing `puts "no indentation spaces"`, and running `gofmt` on it at the end?

I think it's probably a good idea. we in fact do run gofmt when generating the code in stellar-go. However, that requires a bigger refactoring

2opremio:
@bartekn PTAL

lib/xdrgen/generators/go.rb
paulbellamy: Either way, probably outside the scope of this PR.

paulbellamy (approved):

bartekn (approved):
Awesome!

lib/xdrgen/generators/go.rb
```diff
@@ -580,8 +815,16 @@ def render_top_matter(out)
             xdrType()
           }
 
+          type xdrDecodable interface {
```
leighmcculloch, line 818: ❗ Could you add for every type a cast to an unnamed variable? We do that with all the other interfaces in here to make sure all the types satisfy the interfaces. i.e.:

```go
var _ xdrDecodeable = (*#{name})(nil)
```

leighmcculloch: ❗ Could you name this interface `decoderFrom`? That's the pattern used for similar named interfaces in the stdlib. e.g. The interface that contains the `ReadFrom` function is [ReaderFrom](https://pkg.go.dev/io#ReaderFrom).

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+        name = name(typedef)
+        type = typedef
+        out.puts <<-EOS.strip_heredoc
+        // DecodeFrom decodes this value using the Decoder.
+        func (e *#{name}) DecodeFrom(d *xdr.Decoder) (int, error) {
+          v, n, err := d.DecodeInt()
+          if err != nil {
+            return n, err
+          }
+          if _, ok := #{private_name type}Map[v]; !ok {
```
leighmcculloch, line 587: 💭 Given that most of our enums start at zero and are sequential, and given that use of the map will now be in the hot path, I wonder if for enums that we detect have that property, if we could simply replace this with a `if v >= len(...Map) { // error }`? We don't have to do this, and we definitely don't have to do this in this PR, but maybe something to evaluate if there is a saving here over hitting the map.

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+          end
+        when AST::Typespecs::Simple
+          case type.sub_type
+          when :simple, :optional
+            optional_within = type.is_a?(AST::Identifier) && type.resolved_type.sub_type == :optional
+            if optional_within
+              render_variable_declaration(out, "  ", 'b', "bool", declared_variables: declared_variables)
+              out.puts "  b, nTmp, err = d.DecodeBool()"
+              out.puts tail
+              out.puts "  #{var} = nil"
```
leighmcculloch, line 685: 🎉 I really like that this sets the value of every field.

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+      end
+
+      # render_decode_from_body assumes there is an `d` variable containing an
+      # xdr.Decoder, and a variable defined by `var` that is the value to
+      # encode.
+      def render_decode_from_body(out, var, type, declared_variables:, self_encode:)
+        tail = <<-EOS.strip_heredoc
+          n += nTmp
+          if err != nil {
+            return n, err
```
leighmcculloch, line 637: 💡 It may be helpful to debugging and understanding future xdr decoding failures if we annotate the error with the name of the type being decoded. The type appears to be known here so I think this is a small change for a possible huge benefit?

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+            mn = name(arm)
+            type = arm.type
+            out2.puts "  u.#{mn} = new(#{reference arm.type})"
+            render_decode_from_body(out2, "(*u.#{mn})",type, declared_variables: [], self_encode: false)
+          end
+          out2.puts "  return n, nil"
+          out2.string
+        end
+        unless union.default_arm.present?
+          out.puts "  return n, fmt.Errorf(\"#{name(union.discriminant)} (#{reference union.discriminant.type}) switch value '%d' is not valid for union #{name}\", u.#{name(union.discriminant)})"
```
leighmcculloch, line 571: ❗ This error message looks a bit confusing once rendered. e.g.:
```go
fmt.Errorf("Type (EnvelopeType) switch value '%d' is not valid for union TransactionEnvelope", u.Type)
```
I suggest changing the order of the components so the error provides context first then zooms in:
```go
fmt.Errorf("union TransactionEnvelope has invalid Type (EnvelopeType) switch value '%d'", u.Type)
```
I think this can be achieved with:
```suggestion
          out.puts "  return n, fmt.Errorf(\"union #{name} has invalid #{name(union.discriminant)} (#{reference union.discriminant.type}) switch value '%d'\", u.#{name(union.discriminant)})"
```

lib/xdrgen/generators/go.rb
```diff
@@ -549,6 +531,259 @@ def render_encode_to_body(out, var, type, self_encode:)
...
+            if optional_within
+              render_variable_declaration(out, "  ", 'b', "bool", declared_variables: declared_variables)
+              out.puts "  b, nTmp, err = d.DecodeBool()"
+              out.puts tail
+              out.puts "  #{var} = nil"
+              out.puts "  if b {"
+              out.puts "     #{var} = new(#{name type.resolved_type.declaration.type})"
+            end
+            var = "(*#{name type})(#{var})" if self_encode
+            out.puts "  nTmp, err = #{var}.DecodeFrom(d)"
```
leighmcculloch, line 690: ❔  I noticed in stellar/go#4064 that this generates the following code where `SponsoringId` is a `type SponsorshipDescriptor *AccountId`:
```go
		s.SponsoringId = new(AccountId)
		nTmp, err = s.SponsoringId.DecodeFrom(d)
```
I'm surprised you can call `DecodeFrom` on a `SponsorshipDescriptor` since that type has no functions. I would have thought you'd need to either dereference `s.SponsoringId` first so that you have an `AccountId` or convert it to `*AccountId`.

I see stellar/go#4064 passed its tests though, so I assume it compiles and works. I don't understand why though because `SponsorshipDescriptor` shouldn't inherit any functions.

If you know why, please share.

leighmcculloch (approved):
Three requests (❗) otherwise LGTM. I definitely don't understand all of it, unfortunately xdrgen is pretty complex, but the areas of the generated code I sampled in stellar/go#4064 look good.

I also left some non-blocking comments: a suggestion (💡), a thought for future improvements that may or may not be beneficial (💭), and a question about something I don't understand (❔).

🔍 I see above we ran a verify range test with 100k ledgers. That's only 5 days worth of ledgers. Given the larger surface area of xdr decoding and its risk for Horizon, and that Horizon supports reingestion of much larger periods of time, should we perform a verify range of a much larger range of time before merging? cc @bartekn

lib/xdrgen/generators/go.rb
2opremio: Makes sense, I will also add it for `EncodeTo`

2opremio: Will do

leighmcculloch: +1 As a follow up we can remove the xdrType thing I added too, it's superfluous and wasn't the best approach I took. I'm happy to do the clean up for EncodeTo, but if you're going to do it 👍🏻 sgtm.

lib/xdrgen/generators/go.rb
2opremio: Good idea, However:

There are many enums with negative values. Some examples are: `BucketEntryType`, `CreateAccountResultCode`, `PaymentResultCode`, `PathPaymentStrictReceiveResultCode` ....

Not even a range is generally a possibility since the XDR definition can contain scattered values.

We could infer a range where available and use that, but I think it is an overoptimization (without further proof of a bottleneck) and definitely out of the scope of this PR.

Let's later see if further profiling reveals map access as a problem.

lib/xdrgen/generators/go.rb
2opremio: Good point, I will incorporate it (and add it in a later PR for encoding)

lib/xdrgen/generators/go.rb
2opremio: I was mimicking the error in go-xdr, but sure, I will change it

lib/xdrgen/generators/go.rb
2opremio: If you could cleanup the encodeto part (adding errors, making it consistent with this etc ...). I would appreciate it.

lib/xdrgen/generators/go.rb
2opremio: Uhm, good point. I think it just uses the `*AccountId` method.

leighmcculloch: How though. `SponsorshipDescriptor` is its own type, it doesn't inherit methods. The type of the value can be converted back to *AccountId, or it can magically become it through dereferencing to `AccountId`, but I don't see how this works. 🤔 

For example, the following code fails to compile.

```go
package main

type AccountId int

func (a AccountId) Method() {}

type SponsorshipDescriptor *AccountId

func main() {
	var s SponsorshipDescriptor
	s = new(AccountId)
	s.Method()
}
```
Ref: https://play.golang.org/p/kTxwVGPMe4X

As far as I can tell the type of `s.SponsoringId` is `SponsorshipDescriptor`.

2opremio: Your example is not accurate, it's missing an equal, Sponsoring descriptor is an alias.

See https://play.golang.org/p/LBAb4oxe2vm

2opremio: See 

```
        if typedef.sub_type == :optional
          out.puts "type #{name typedef} = #{reference typedef.declaration.type}"
        else
          out.puts "type #{name typedef} #{reference typedef.declaration.type}"
        end
```

leighmcculloch: Of course! Haha, [I was the one who changed it to be an alias](https://github.com/stellar/xdrgen/commit/92a28e1f3ef3a2b7e9409a2baa2c5ed0751d3e1c#diff-559f5809716dacbc678a75e354ead997243db1419c9186a359d5ebf749fed6e2R28-R31) and I forgot 🤦🏻.

PR #72, opened 2021-11-12
Add multiple EncodeTo improvements

### LINKED ISSUE #70: `EncodeTo()` should error when encountering invalid union discriminant values
xdrgen generates the following Go code:

```go
func (s StellarValueExt) EncodeTo(e *xdr.Encoder) error {
	_, err := e.EncodeInt(int32(s.V))
	if err != nil {
		return err
	}
	switch StellarValueType(s.V) {
	case StellarValueTypeStellarValueBasic:
		// Void
	case StellarValueTypeStellarValueSigned:
		err = (*s.LcValueSignature).EncodeTo(e)
		if err != nil {
			return err
		}
	}
	return err
}
```

for ext field of:

```
/* StellarValue is the value used by SCP to reach consensus on a given ledger
 */
struct StellarValue
{
    Hash txSetHash;      // transaction set to apply to previous ledger
    TimePoint closeTime; // network close time

    // upgrades to apply to the previous ledger (usually empty)
    // this is a vector of encoded 'LedgerUpgrade' so that nodes can drop
    // unknown steps during consensus if needed.
    // see notes below on 'LedgerUpgrade' for more detail
    // max size is dictated by number of upgrade types (+ room for future)
    UpgradeType upgrades<6>;

    // reserved for future use
    union switch (StellarValueType v)
    {
    case STELLAR_VALUE_BASIC:
        void;
    case STELLAR_VALUE_SIGNED:
        LedgerCloseValueSignature lcValueSignature;
    }
    ext;
};
```

The code should not allow invalid values of `StellarValueExt.V`

2opremio:
Enum econding:
  * Check that enum values are legal before encoding them
  * Use `e` as the receiver name of `EncodeTo` (it was using `s`) which is inconsistent with the rest of the code.
  
Union encoding:
  * Error when a supplied discriminant doesn't match a branch ( #70 )
  * When an enum is used as a discrimimnant, check that they are legal before encoding
  * Use `u` as the receiver name of `EncodeTo` (it was using `s`) which is inconsistent with the rest of the code.

Layout improvements:
  * Factor invocations and error checking. This makes the code more compact and removes error [S1021](https://staticcheck.io/docs/checks#S1021 from the generated code:
```diff
-       err = s.SeqNum.EncodeTo(e)
-       if err != nil {
+       if err = s.SeqNum.EncodeTo(e); err != nil {
                return err
        }
```
  * Fix incorrectly nested error checks
```diff
        if s.InflationDest != nil {
-               err = (*s.InflationDest).EncodeTo(e)
-       }
-       if err != nil {
-               return err
+               if err = (*s.InflationDest).EncodeTo(e); err != nil {
+                       return err
+               }
        }
```

Closes #70

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -353,22 +353,26 @@ module Xdrgen
       def render_union_encode_to_interface(out, union)
         name = name(union)
         out.puts "// EncodeTo encodes this value using the Encoder."
-        out.puts "func (s #{name}) EncodeTo(e *xdr.Encoder) error {"
-        out.puts "  _, err := e.EncodeInt(int32(s.#{name(union.discriminant)}))"
-        out.puts "  if err != nil {"
-        out.puts "    return err"
-        out.puts "  }"
-        switch_for(out, union, "s.#{name(union.discriminant)}") do |arm, kase|
+        out.puts "func (u #{name}) EncodeTo(e *xdr.Encoder) error {"
+        out.puts "  var err error"
+        render_encode_to_body(out, "u.#{name(union.discriminant)}", union.discriminant.type, self_encode: false)
+        switch_for(out, union, "u.#{name(union.discriminant)}") do |arm, kase|
           out2 = StringIO.new
           if arm.void?
-            "// Void"
+            out2.puts "// Void"
           else
             mn = name(arm)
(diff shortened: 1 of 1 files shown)


2opremio:
You can see the generated code at https://github.com/stellar/go/pull/4068

bartekn (approved):

PR #77, opened 2022-04-19
Add Python generation support. 

overcat:
This is the version the Python SDK is currently using, you can find the generated files [here](https://github.com/StellarCN/py-stellar-base/tree/3dec99b11a06df87dd4914c65068e6916afbc6e3/stellar_sdk/xdr).


overcat:
I plan to keep it in sync with master, so convert it to draft first

lib/xdrgen/generators/python.rb
```diff
@@ -0,0 +1,666 @@
...
+            union.arms.each do |arm|
+              next if arm.void?
+              render_array_length_checker arm, out
+            end
+
+            out.puts "self.#{union_discriminant_name_underscore} = #{union_discriminant_name_underscore}"
+            union.arms.each do |arm|
+              next if arm.void?
+              arm_name_underscore = arm.name.underscore
+              out.puts "self.#{arm_name_underscore} = #{arm_name_underscore}"
```
leighmcculloch, line 202: These few lines appear to work only on arms, and not arm cases, which means cases that are grouped together will produce only a single line here. Is that as intended?

leighmcculloch (commented):
Thanks for upstreaming this code generator!

lib/xdrgen/generators/python.rb
overcat: I think it's working fine, it is used to generate the following code:

https://github.com/StellarCN/py-stellar-base/blob/560519d2821746ba6ae6164bcf5a471744066cbe/stellar_sdk/xdr/ledger_entry_change.py#L39-L42

leighmcculloch: Yes that looks good. Also, TransactionResultResult also looks good and is a more complex union definition. https://github.com/StellarCN/py-stellar-base/blob/560519d2821746ba6ae6164bcf5a471744066cbe/stellar_sdk/xdr/transaction_result_result.py

.github/workflows/ruby.yml
```diff
@@ -32,5 +32,13 @@ jobs:
...
     - name: Run Rspec tests
       run: bundle exec rspec
+    - name: Format Python Output - Set up Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.9'
+    - name: Format Python Output - Install dependencies
+      run: pip install autoflake==1.4 black==22.3.0
+    - name: Format Python Output - Format
+      run: autoflake --in-place --ignore-init-module-imports --remove-all-unused-imports spec/output/generator_spec_python/*/*.py && black spec/output/generator_spec_python
```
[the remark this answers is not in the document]
overcat: I think formatted code is easier for humans to review. 

We don't enforce what formatting tools should be used, users can make their own choices.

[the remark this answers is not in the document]
leighmcculloch: I agree, but anyone updating the xdrgen code needs pip, autoflake, and black installed to update the test files for the tests to pass, which is not ideal. Does the code work and is it readable without these? If so, I think it'd fine to remove them for the tests. People will and can still use them when using xdrgen to generate code.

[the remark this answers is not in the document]
overcat: Agree, fixed.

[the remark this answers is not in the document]
leighmcculloch: With formatting removed in 27f2a27da8a350766433be3845a0eec43ae68cf3 I think it looks fine, not as good as the formatted, but for the purpose of the tests and seeing what the raw generated code is it is fine. 👍🏻

leighmcculloch (approved):
Thanks @overcat! This looks great and nice and feature complete too with support for default arms and length checks.

@accordeiro you have more experience with Python than I do, would you be able to 👀 this change as well? If it looks good we can merge.

leighmcculloch:
> @accordeiro you have more experience with Python than I do, would you be able to 👀 this change as well? If it looks good we can merge.

@JakeUrban Would you be able to review this PR?

lib/xdrgen/generators/python.rb
```diff
@@ -0,0 +1,666 @@
...
+          out.puts "def unpack(cls, unpacker: Unpacker) -> \"#{typedef_name}\":"
+          out.indent(2) do
+            decode_member typedef, out
+            out.puts "return cls(#{typedef_name_underscore})"
+          end
+          render_xdr_utils(out, typedef_name)
+          out.puts <<~HEREDOC
+            def __eq__(self, other: object):
+                if not isinstance(other, self.__class__):
+                    return NotImplemented
```
Shaptic, line 138: is `raise` more appropriate here?

Shaptic (approved):
not sure why we haven't merged this yet but we should

lib/xdrgen/generators/python.rb
overcat: We should use `return NotImplemented` here, please refer to [Python Docs](https://docs.python.org/3.11/library/constants.html#NotImplemented).

> A special value which should be returned by the binary special methods (e.g. [__eq__()](https://docs.python.org/3.11/reference/datamodel.html#object.__eq__), [__lt__()](https://docs.python.org/3.11/reference/datamodel.html#object.__lt__), [__add__()](https://docs.python.org/3.11/reference/datamodel.html#object.__add__), [__rsub__()](https://docs.python.org/3.11/reference/datamodel.html#object.__rsub__), etc.) to indicate that the operation is not implemented with respect to the other type; may be returned by the in-place binary special methods (e.g. [__imul__()](https://docs.python.org/3.11/reference/datamodel.html#object.__imul__), [__iand__()](https://docs.python.org/3.11/reference/datamodel.html#object.__iand__), etc.) for the same purpose. It should not be evaluated in a boolean context. NotImplemented is the sole instance of the [types.NotImplementedType](https://docs.python.org/3.11/library/types.html#types.NotImplementedType) type.

leighmcculloch:
> not sure why we haven't merged this yet but we should

> @JakeUrban Would you be able to review this PR?

lib/xdrgen/generators/python.rb
Shaptic: nice, thank you for the reference! :pray:

PR #79, opened 2022-04-22
Add Rust generator

### LINKED ISSUE #76: Merge in downstream Rust generator
Now that more folks are writing Rust code that interacts with Stellar, including us with [rs-stellar-contract-sdk](https://github.com/stellar/rs-stellar-contract-sdk), there would be a benefit to maintaining a Rust generator inside xdrgen.

There is a Rust generator implemented in a fork of xdrgen: [aurora-rs/xdrgen](https://github.com/aurora-rs/xdrgen).
The generator is currently being used by the [aurora-rs/stellar-base-rs](https://github.com/aurora-rs/stellar-base-rs) repo.
The generator generates types from .x definitions, and relies on the `xdr-rs-serialize` crate for the serialization.

It looks like @fracek is the author of the commits at that fork that add Rust support. @fracek also authored the C# generator that was added into xdrgen in #58 / #59.

cc @graydon

leighmcculloch:
### What
Add a Rust generator.

### Why
While developing [rs-stellar-contract-sdk] @graydon pointed out that we are reimplementing XDR types in the mock host, and it would be better if we were using the XDR types.

Close #76
Close stellar/rs-stellar-contract-sdk#41
Close stellar/rs-stellar-xdr#14

### Notes
This implementation is dependent only on the `base64` crate. It has no dependencies on any other XDR serialization crate and is otherwise self-contained. A cursory review of other XDR crates showed up missing features or crates that had slightly different goals, such as bundling xdr+json encoding. Also, some crates were macro heavy which can make debugging difficult. On the contrary this generator produces 100% macro free code that is easy to navigate and debug.

Fixed length arrays in this generator are represented as fixed length arrays in Rust. This has some downsides, and we may want to revisit that decision. Revisiting that decision will require changing how the library handles length validation.

### Completeness
The generator is feature equivalent to the Go generator when used with the Stellar protocol XDR. Default arms are technically supported by the Go implementation, but not the Rust implementation. Floats are technically supported by the Go implementation, and not by the Rust implementation. Both of these deficits are not used by the Stellar protocol.

This generator needs more testing, and so warnings are included noting it is experimental.

### Special Thanks
Special thanks to @fracek. The first commit of this PR is upstreaming @fracek's Rust generator from https://github.com/aurora-rs/xdrgen/commit/20ae1636da29bfc30dbe3604ee9c25856b21a3ac, which formed the initial basis of this PR.

[rs-stellar-contract-sdk]: https://github.com/stellar/rs-stellar-contract-sdk


leighmcculloch:
TODOs for this are being tracked at:
https://github.com/stellar/rs-stellar/xdr

leighmcculloch:
This code generator is complete enough I'd like to merge it now, and iterate on it further. It is not sufficiently tested yet, but I've added a warning that prints if you use it noting it is experimental at this stage.

leighmcculloch:
@bartekn @graydon @MonsieurNicolas Could you review and 👍🏻 ?

MonsieurNicolas:
I think from a testing point of view it's fine to not have exhaustive testing, but maybe adding a "sample expected output" based on some xdr would allow to act as a "smoke test" for it as well as get a rough idea of how the generated code looks like.

I don't have much of an opinion if following what go does is a good idea or not as I don't know what rustaceans expect.

spec/output/generator_spec_rust/union.x/MyXDR.rs
```diff
@@ -0,0 +1,779 @@
...
+        }
+
+        impl ReadXdr for IntUnion {
+            #[cfg(feature = "std")]
+            fn read_xdr(r: &mut impl Read) -> Result<Self> {
+                let dv: i32 = <i32 as ReadXdr>::read_xdr(r)?;
+                #[allow(clippy::match_same_arms, clippy::match_wildcard_for_single_variants)]
+                let v = match dv {
+                    0 => Self::V0(i32::read_xdr(r)?),
+1 => Self::V1(VecM::<i32>::read_xdr(r)?),
```
graydon, line 720: Minor issue but these seem mis-indented

graydon (approved):
As near as I can tell it looks good!

spec/output/generator_spec_rust/union.x/MyXDR.rs
leighmcculloch: It might be possible for me to fix this a little, but we don't generally try to make xdrgen produce formatted code, and instead we run its output through the languages code formatter outside of xdrgen.

leighmcculloch:
Upon reviewing this myself today I realized I missed a check on pads, we should be checking that any padded bytes are zero bytes. And, I also missed even adding padding to the fixed-length opaque data types. The reason I missed this is because the stellar-protocol tx types don't need padding on any of its fixed-length types, since any fixed length values are already the correct length. I've added TODOs, and will fix that, then merge.

PR #80, opened 2022-04-27
Add makefile

leighmcculloch:
### What
Add Makefile.
### Why
For conveniently running tests locally.


acharb (approved):
I'm not too familiar with this repo so I would defer to @graydon , but the command lgtm!

PR #84, opened 2022-06-01
Add test output checks to CI

leighmcculloch:
### What
Add test output checks to CI that check that the outputs of the test XDR hasn't changed.

### Why
We have tests that generate output, but there is no assertions on them other than checking if the application encounters an error. We could commit the output which would make PR diffs better since PR diffs would include a snapshot of what will change in generated output.

This was an idea suggested by @MonsieurNicolas in https://github.com/stellar/xdrgen/pull/79#issuecomment-1142886058.


MonsieurNicolas (approved):
Very cool thank you

PR #90, opened 2022-06-15
Rust: Add Deref<Target=Vec<_>> impls for VecMs

leighmcculloch:
### What
Add to the Rust generated code `Deref<Target=Vec<_>>` implementations for `VecM` and all types that are typedefs to `VecM`.

### Why
`Vec` in the stdlib has `[T]` as its Deref which is convenient, and `VecM` could do the same also for convenience. It derefs to `Vec<T>` since `[T]` comes automatically since they're automatically chained.


graydon (approved):
LGTM!

PR #96, opened 2022-06-17
Rust: Add VecM<>str conversions

leighmcculloch:
### What
Add VecM<>str conversions to the Rust generated code.

### Why
VecM's are used for XDR strings.


graydon (approved):
Super helpful, thanks!

PR #99, opened 2022-06-18
Rust: Fix to_vec and add to_string

leighmcculloch:
### What
Make to_vec not consume the vec and create a new vec. Add to_string to do same as vec but for a string.

### Why
to_ functions should not consume an owned type unless they are Copy. The to_string function is helpful because we use VecM's for strings, and it is not always convenient to try_into a string.

### DIFF
--- a/lib/xdrgen/generators/rust/src/types.rs
+++ b/lib/xdrgen/generators/rust/src/types.rs
@@ -416,13 +416,23 @@ impl<T, const MAX: u32> VecM<T, MAX> {
     }
 
     #[must_use]
-    pub fn to_vec(self) -> Vec<T> {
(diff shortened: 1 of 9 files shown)


graydon (approved):
I don't really care either way but I thought _to or _into (at least one of them?) ought to be self-consuming.

leighmcculloch:
to_ shouldn't, according to convention, consume unless it is a Copy type. But you're right, we could have an into_ that consumes. Added in cc2bc93.

Ref: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

PR #100, opened 2022-06-21
Rust: Add name fn to enums, unions

leighmcculloch:
### What
Add name fn to enums and unions that returns a string representation of the case.

### Why
@graydon mentioned that it is needed in the host.


graydon (approved):
LGTM, thanks!

PR #102, opened 2022-06-22
Rust: add read_xdr_iter

leighmcculloch:
### What
Add a read_xdr_iter to all ReadXdr impls that returns an iterator for reading a stream of values from a Read impl.

### Why
In a couple places we need to read a stream of XDR values which is pretty inconvenient to impl in practice. Wrapping it into an iterator should make it much simpler.

### Also
This highlights a problem with existing read_xdr impls that they use read_exact which unfortunately obscures partial fill EOFs which could be a security concern in our uses. A TODO has been added to address this later.

### DIFF
--- a/lib/xdrgen/generators/rust/src/types.rs
+++ b/lib/xdrgen/generators/rust/src/types.rs
@@ -124,7 +124,7 @@ impl<'r, R: Read, S: ReadXdr> ReadXdrIter<'r, R, S> {
 impl<'r, R: Read, S: ReadXdr> Iterator for ReadXdrIter<'r, R, S> {
     type Item = Result<S>;
 
-    // Next reads the internal reader and XDR decoded it into the Self type. If
+    // Next reads the internal reader and XDR decodes it into the Self type. If
     // the EOF is reached without reading any new bytes `None` is returned. If
     // EOF is reached after reading some bytes a truncated entry is assumed an
     // an `Error::Io` containing an `UnexpectedEof`. If any other IO error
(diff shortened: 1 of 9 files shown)


lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -101,6 +104,39 @@ impl From<Error> for () {
...
+
+#[cfg(feature = "std")]
+impl<'r, R: Read, S: ReadXdr> Iterator for ReadXdrIter<'r, R, S> {
+    type Item = Result<S>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        match S::read_xdr(&mut self.r) {
+            Ok(s) => Some(Ok(s)),
+            // TODO: Distinguish between EOF and EOF-with-a-partial-fill. An EOF
+            // that occurs at teh end of an item indicates the end of a stream
```
graydon, line 128: teh end

graydon (approved):
I don't exactly understand the buffering regime but it seems plausible and if it works I'm happy to have more ways to read stuff!

leighmcculloch:
> I don't exactly understand the buffering regime but it seems plausible and if it works I'm happy to have more ways to read stuff!

Any reader should probably be buffered anyway, but ideally control over that is in the hands of the user and ideally we aren't creating a BufReader internally. The reason I'm doing it here is it's a quick way to get around the read_exact limitations on EOF errors. When reading a stream in the iterator I need to peak at the next byte to see if it is an EOF error. If it is then we have a clean termination of the stream. If read_exact sees an EOF then that's an unclean termination of the stream. read_exact on its own doesn't let us distinguish these situations for decent reasons.

An alternative to using the BufReader would be to define our own reader that read just the next byte, and then on the next call it got to its read function it'd return that byte if it had it, otherwise pass through to the underlying reader. This might be simpler? Maybe less efficient, not sure. It's essentially the same thing as BufReader but a single byte instead. This is all internal so we can change this if needed without changing the API.

leighmcculloch:
Actually, I probably should circle back and replace it with the PeekRead, or remove use of read_exact... because otherwise people will have double nested BufReader's ugh.

leighmcculloch:
Will follow up in https://github.com/stellar/xdrgen/issues/103.

Issue #107, opened 2022-06-25
Rust: name() fns should return &'static str


graydon:
Returning &str means we can't use it in a function where we construct-and-throw-away an enumeration, we have to construct it outside to live as long as the str it returns. Pretty sure this is an oversight.

Issue #106, opened 2022-06-25
Rust: discriminant(), name() fns should be const


graydon:
Currently this prevents using it in const fns. As far as I know, it is logically const.

PR #109, opened 2022-07-05
Rust: Make name, discriminant const and trait fns

### LINKED ISSUE #108: Rust: generate trait that covers access-to-the-name-of-an-enum such that one can write generic code against all enums.
Ideally same for unions and any other types that have common `impl` blocks. Abstracting over all-types-with-the-same-impl-block is what traits are for!

leighmcculloch:
### What
In the Rust generator make the `name` and `discriminant` fns on enums and unions `const` and add `trait`s that group these fns. Also changed the `&str` return values so that they have a `'static` lifetime.

### Why
The fns can be const fns which increases the places we can use them, and their str values are easier to use if they have a static lifetime. @graydon asked for traits so it is easier to write code across all enum / union types.

Close #108 #107 #106 

### Additional Thoughts
Now that we're adding traits to the generated code it is becoming more interesting to have a use case where these traits could be usable across multiple sets of generated code which would mean moving out some of the code in `types.rs` into its own crate and repo. We may like to do this down the track.

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -182,7 +182,7 @@ module Xdrgen
         out.puts <<-EOS.strip_heredoc
         impl #{name enum} {
             #[must_use]
-            pub const fn name(&self) -> &str {
+            pub const fn name(&self) -> &'static str {
                 match self {
                     #{enum.members.map do |m|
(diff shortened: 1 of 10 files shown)


graydon (approved):
I think this handles the issues, yeah. Will have to try, but LGTM!

Issue #112, opened 2022-07-07
Rust: Add JSON ser/der to all generated types


leighmcculloch:
We should make it easy to JSON serialize/deserialize the XDR types. This is somewhat controversial as there is no canonical format for JSON-to-XDR, so this would just be a Rust flavored JSON representation of the XDR for non-critical data use, and without any backwards compatible guarantees.

This should be added behind a feature flag so that none of the JSON functionality is included by default.

@sisuresh did some experimentation with serde_json and found it worked well: 

https://github.com/sisuresh/rs-stellar-xdr/commit/ebfd6495f732ff39e8c8371c28cf566222dc1dca#diff-20a7ada4b13b7b258fd3b69d364ba4a655f9988488e5f890094e75412b8591c6

We have to make a choice to flavor the resulting JSON fields names to be consistent with the Rust types, or consistent with the XDR. Both options are possible. The former is more aligned with the first paragraph of this issue, but the latter might make more sense too for any user of the JSON who is comparing it to the XDR. I'm inclined to lean towards the latter.

leighmcculloch:
I'm going to close this because this was done in #117, and separate issues exist now for follow work:
- https://github.com/stellar/xdrgen/issues/118
- https://github.com/stellar/xdrgen/issues/119

PR #117, opened 2022-07-12
Rust: Add serde json support

sisuresh:
Initial implementation of #112.

Deserialization of `Box` and `VecM` is not implemented yet without `alloc`. Should be fixed as part of https://github.com/stellar/rs-stellar-xdr/issues/47.

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -134,7 +134,7 @@ module Xdrgen
 
       def render_struct(out, struct)
         out.puts "#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]"
-        out.puts "#[cfg_attr(all(feature = \"serde\", feature = \"alloc\"), derive(Serialize, Deserialize), serde(rename_all = \"camelCase\"))]"
+        out.puts %{#[cfg_attr(all(feature = "serde", feature = "alloc"), derive(serde::Serialize, serde::Deserialize), serde(rename_all = "camelCase"))]}
         out.puts "pub struct #{name struct} {"
         out.indent do
           struct.members.each do |m|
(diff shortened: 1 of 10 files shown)


lib/xdrgen/generators/rust.rb
```diff
@@ -134,6 +134,7 @@ def render_source_comment(out, defn)
 
       def render_struct(out, struct)
         out.puts "#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]"
+        out.puts "#[cfg_attr(feature = \"serde\", derive(Serialize, Deserialize), serde(rename_all = \"camelCase\"))]"
```
leighmcculloch, line 137: Can you gate the serde code on alloc too?
```suggestion
        out.puts "#[cfg_attr(all(feature = \"serde\", feature = \"alloc\"), derive(Serialize, Deserialize), serde(rename_all = \"camelCase\"))]"
```

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -1,5 +1,8 @@
 use core::{array::TryFromSliceError, fmt, fmt::Debug, ops::Deref};
 
+#[cfg(feature = "serde")]
+use serde::{Deserialize, Serialize};
```
leighmcculloch, line 4: I'd be inclined to remove the `use` and fully qualify the serde types at their call sites? I don't think it will be too burdensome, and I think it will aid readability since the type names are rather generic and the crate name will help provide context at their use.

lib/xdrgen/generators/rust.rb
leighmcculloch, line 137: It is common in Ruby to use alternative string style when the string will contain double quotes:
```suggestion
        out.puts %{#[cfg_attr(feature = "serde", derive(Serialize, Deserialize), serde(rename_all = "camelCase"))]}
```

leighmcculloch (commented):
Couple of small things, otherwise this looks good. Before we merge this I'd like to experiment with the output in rs-stellar-xdr to see if there are other things we should tweak.

lib/xdrgen/generators/rust/src/types.rs
sisuresh: Done

lib/xdrgen/generators/rust.rb
sisuresh: Done

leighmcculloch:
Some other things I'd like to explore, but depending on difficulty we can kick them down the road:

 - VecM's all being serialized as base64.
 - Fixed size arrays being serialized as hex.

leighmcculloch (approved):

PR #127, opened 2022-08-05
Rust: Support Arbitrary on all generated types

graydon:
Sketch of code to generate https://github.com/stellar/rs-stellar-xdr/pull/132 -- support [Arbitrary](https://crates.io/crates/arbitrary) on all XDR types in Rust. Arbitrary is the standard interface to fuzzers and at least one property-testing library [arbtest](https://crates.io/crates/arbtest)

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -133,7 +133,7 @@ module Xdrgen
       end
 
       def render_struct(out, struct)
-        out.puts "#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]"
+        out.puts "#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord, Arbitrary)]"
         out.puts %{#[cfg_attr(all(feature = "serde", feature = "alloc"), derive(serde::Serialize, serde::Deserialize), serde(rename_all = "camelCase"))]}
         out.puts "pub struct #{name struct} {"
         out.indent do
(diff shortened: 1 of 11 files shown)


leighmcculloch:
Could we put arbitrary behind a feature with the same name? I realize that will increase the feature powerset, but it shouldn't impact test and build times as we group all dependency features into a single build.

leighmcculloch:
> property-testing library `arbtest`

Is this compatible with `proptest`? It is very popular.

graydon:
> Is this compatible with proptest? It is very popular.

I was going to say "no, proptest has a completely different interface" but this is obviously an annoying answer, so I present:

https://crates.io/crates/proptest-arbitrary-interop

leighmcculloch:
I think it might make sense for us to implement `proptest::strategy::Strategy` directly in xdrgen, so that people don't need to know how to mix Arbitrary + proptest, but that doesn't need to block this PR.

graydon:
The way `Strategy` is expected to be used in `proptest` is less with a canonical strategy per type, and more with a selection of multiple strategy-returning functions per type, each generating a different set of them. There's a strategy called `any<T>` that hooks into a canonical-strategy-for-a-type (and which I'm imitating here with `arb<T>`) but that's not the only `Strategy` you're expected to use very often; usually people would generate with something more specific to the relationships they expect to hold in a test, like some composition of base types, ranges, tuples, `select`, `sample`, `filter` and such.

We could implement proptest's canonical-per-type strategy adaptor, `proptest::arbitrary::Arbitrary` in addition to `arbitrary::Arbitrary`, but .. the derive for it is unstable & unmaintained currently, and there'd be a lot more code to write it manually, so I figured the adaptor I wrote yesterday to `arbitrary::Arbitrary` should be good enough.

graydon:
LGTM!

leighmcculloch (approved):

PR #130, opened 2022-08-16
Add VecM::into_opton, and related fns

leighmcculloch:
### What
Add VecM::into_opton, and related fns.

### Why
So that converting from single max-length VecM's to Option's is convenient.

Convenience function for work I'm currently doing for https://github.com/stellar/rs-soroban-sdk/issues/118.


paulbellamy (approved):
LGTM, but should we do a "from_option" as well for symmetry?

leighmcculloch:
> should we do a "from_option" as well for symmetry

Oh that's a good idea. I don't need it right now, but we can add it when we need it.

PR #132, opened 2022-08-18
Add Gemfile.lock, update rake to 12.*

leighmcculloch:
### What
Add Gemfile.lock, update rake to 12.*, and a Makefile target for convenience access to a Ruby installation with the working directory..

### Why
Xdrgen is an application and we should be committing a Gemfile.lock so that the dependencies we use to build it during development are stable, and so we have full control over what they are.

Update rake to 12.* to get the latest version.

The Makefile change is just for convenience.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 /.bundle/
 /.yardoc
-/Gemfile.lock
 /_yardoc/
 /coverage/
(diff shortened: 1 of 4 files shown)


C0x41lch0x41 (commented):
LGTM thanks!

PR #135, opened 2022-09-05
Go: Misc code generation fixes

2opremio:
1. Fix the generation of file hashes
2. Make sure we only generate EncodeTo by pointer for fixed opaque types or fixed opaque synonyms

### DIFF
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -410,7 +410,7 @@ module Xdrgen
         out.puts "// EncodeTo encodes this value using the Encoder."
         if is_fixed_array_type(type) ||
             (type.is_a?(AST::Identifier) && type.sub_type == :simple && type.resolved_type.is_a?(AST::Definitions::Typedef) && is_fixed_array_type(type.resolved_type.declaration.type))
-          # Implement EncodeTo by pointer in for Go array types
+          # Implement EncodeTo by pointer for Go array types
           # otherwise (if called by value), Go will make a heap allocation
           # for every by-value call since the copy required by the call
           # tends to escape the stack due to the large array sizes.
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -409,7 +409,7 @@ module Xdrgen
         type = typedef.declaration.type
         out.puts "// EncodeTo encodes this value using the Encoder."
         if is_fixed_opaque(type) ||
-           (type.is_a?(AST::Identifier) && type.resolved_type.is_a?(AST::Definitions::Typedef) && is_fixed_opaque(type.resolved_type.declaration.type))
+           (type.is_a?(AST::Identifier) && typedef.declaration.is_a?(AST::Declarations::Simple) && type.resolved_type.is_a?(AST::Definitions::Typedef) && is_fixed_opaque(type.resolved_type.declaration.type))
           # Implement EncodeTo by pointer in fixed opaque types (arrays)
           # otherwise (if called by value), Go will make a heap allocation
           # for every by-value call since the copy required by the call
--- a/lib/xdrgen/generators/go.rb
+++ b/lib/xdrgen/generators/go.rb
@@ -400,17 +400,17 @@ module Xdrgen
         EOS
       end
 
-      def is_fixed_opaque(type)
-        type.is_a?(AST::Typespecs::Opaque) && type.fixed?
+      def results_in_go_array(type)
+        (type.is_a?(AST::Typespecs::Opaque) && type.fixed?) || type.sub_type == :array
       end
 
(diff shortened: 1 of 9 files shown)


2opremio:
@leighmcculloch Are the sample generated files being tested (i.e. compiled)? I doubt they can work without the ending comma.

spec/output/generator_spec_go/test.x/MyXDR_generated.go
```diff
@@ -411,7 +411,7 @@ var _ xdrType = (*Hash)(nil)
 //
 type Hashes1 [12]Hash
 // EncodeTo encodes this value using the Encoder.
-func (s *Hashes1) EncodeTo(e *xdr.Encoder) error {
+func (s Hashes1) EncodeTo(e *xdr.Encoder) error {
```
2opremio, line 414: This is wrong, we should be using a pointer here. But not when the synonym is a slice

2opremio: Will fix

spec/output/generator_spec_go/test.x/MyXDR_generated.go
```diff
@@ -547,7 +547,7 @@ var _ xdrType = (*Hashes2)(nil)
 //
 type Hashes3 []Hash
 // EncodeTo encodes this value using the Encoder.
-func (s *Hashes3) EncodeTo(e *xdr.Encoder) error {
+func (s Hashes3) EncodeTo(e *xdr.Encoder) error {
   var err error
 if _, err = e.EncodeUint(uint32(len(s))); err != nil {
```
2opremio, line 552: I think we should try to compile the generated files. This doesn't compile.

leighmcculloch: +1

leighmcculloch: 💯  It would be really helpful if we tested the outputs. We should probably do so for the go and rust generated code at least.

lib/xdrgen/generators/go.rb
```diff
@@ -822,7 +822,7 @@ def render_top_matter(out)
         out.puts <<-EOS.strip_heredoc
         // XdrFilesSHA256 is the SHA256 hashes of source files.
         var XdrFilesSHA256 = map[string]string{
-          #{@output.relative_source_path_sha256_hashes.map(){ |path, hash| %{"#{path}": "#{hash}"} }.join(",\n")}
+          #{@output.relative_source_path_sha256_hashes.map(){ |path, hash| %{"#{path}": "#{hash}",} }.join("\n")}
```
leighmcculloch, line 825: ❤️ Thanks for fixing this!

leighmcculloch (approved):
Nice catch. Thanks for fixing these things.

I left some inline comments: suggestions (💡), questions (❔), but regardless 👍🏻.

lib/xdrgen/generators/go.rb
```diff
@@ -400,17 +400,17 @@ def render_enum_encode_to_interface(out, typedef)
...
 
       def render_typedef_encode_to_interface(out, typedef)
         name = name(typedef)
         type = typedef.declaration.type
         out.puts "// EncodeTo encodes this value using the Encoder."
-        if is_fixed_opaque(type) ||
-           (type.is_a?(AST::Identifier) && type.resolved_type.is_a?(AST::Definitions::Typedef) && is_fixed_opaque(type.resolved_type.declaration.type))
-          # Implement EncodeTo by pointer in fixed opaque types (arrays)
+        if results_in_go_array(type) ||
+            (type.is_a?(AST::Identifier) && type.sub_type == :simple && type.resolved_type.is_a?(AST::Definitions::Typedef) && results_in_go_array(type.resolved_type.declaration.type))
```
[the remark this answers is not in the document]
2opremio: It checks that it's a type synonym (from identifier to identifier), or at least that was my intention.

lib/xdrgen/generators/go.rb
2opremio: you're welcome!

lib/xdrgen/generators/go.rb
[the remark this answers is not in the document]
leighmcculloch: Do you need the condition? I'm surprised it's required, because if a resolved type is typedef and the resolved type a fixed length array then it feels appropriate to do this too.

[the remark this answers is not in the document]
2opremio: I do, otherwise the code generates EncodeTo by pointer for array types (as indicated in the commit message)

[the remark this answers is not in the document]
leighmcculloch: Ahh, yes, makes sense.

PR #137, opened 2022-09-16
make xdr hash scheme independent of white spaces

MonsieurNicolas:
On some platforms (Windows, maybe others), .x files may differ only in line ending causing a hash mismatch.
This PR removes all whitespaces when computing hashes as to avoid this problem.

### DIFF
--- a/lib/xdrgen/output.rb
+++ b/lib/xdrgen/output.rb
@@ -18,7 +18,7 @@ module Xdrgen
     end
 
     def relative_source_path_sha256_hashes
-      relative_source_paths.map { |p| [p, Digest::SHA256.file(p).hexdigest] }.to_h
+      relative_source_paths.map { |p| [p, Digest::SHA256.hexdigest(File.read(p).gsub(/[[:space:]]/,''))] }.to_h
     end
 
     def open(child_path)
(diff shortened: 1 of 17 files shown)


lib/xdrgen/output.rb
```diff
@@ -18,7 +18,7 @@ def relative_source_paths
     end
 
     def relative_source_path_sha256_hashes
-      relative_source_paths.map { |p| [p, Digest::SHA256.file(p).hexdigest] }.to_h
+      relative_source_paths.map { |p| [p, Digest::SHA256.hexdigest(File.read(p).gsub(/[[:space:]]/,''))] }.to_h
```
leighmcculloch, line 21: Could we normalize the line endings instead of scrub whitespace from the whole files? Scrubbing whitespace makes it harder to debug and will cause files with space changes to have the same hash.

leighmcculloch: Actually I think https://github.com/stellar/stellar-xdr-next/pull/32 is preferred over this.

MonsieurNicolas:
this is not needed anymore as per https://github.com/stellar/stellar-core/pull/3534

PR #141, opened 2022-09-27
Add consts containing enum variants as strs

leighmcculloch:
### What
Add consts containing enum variants as strs

### Why
So that the values can be used in the soroban-cli.


graydon:
LGTM!

graydon (approved):

PR #142, opened 2022-09-29
Rust: Add read xdr to end fns

leighmcculloch:
### What
Add read xdr to end fns to the generated Rust code, and use them for base64 decoding.

### Why
A lot of the time when reading from a stream into a value we want to read the entire stream into that one value, and for there to be extra data in the stream that is an error, because we don't expect there to be any extra data, and that data might be something we're discarding when we shouldn't be.

It's helpful to have functions that do this for us, so that the caller doesn't have to do this themselves.

The _to_end style functions are aligned with an existing pattern in the std::io::Read trait, where functions that end in _to_end read the stream all the way to the end. Although the function names do read a little odd. There might be a better name for this, but I haven't found it.

### DIFF
--- a/spec/output/generator_spec_rust/block_comments.x/MyXDR.rs
+++ b/spec/output/generator_spec_rust/block_comments.x/MyXDR.rs
@@ -229,9 +229,41 @@ pub trait ReadXdr
 where
     Self: Sized,
 {
+    /// Read the XDR and construct the type.
+    ///
+    /// Read bytes from the given read implementation, decoding the bytes as
+    /// XDR, and construct the type implementing this interface from those
+    /// bytes.
+    ///
+    /// Just enough bytes are read from the read implementation to construct the
+    /// type. Any residual bytes remain in the read implementation.
+    ///
+    /// All implementations should continue if the read implementation returns
+    /// [`ErrorKind::Interrupted`](std::io::ErrorKind::Interrupted).
+    ///
+    /// Use [`ReadXdr::read_xdr_to_end`] when the intent is for all bytes in the
+    /// read implementation to be consumed by the read.
     #[cfg(feature = "std")]
     fn read_xdr(r: &mut impl Read) -> Result<Self>;
 
+    /// Read the XDR and construct the type, and consider it an error if the
+    /// read does not completely consume the read implementation.
+    ///
+    /// Read bytes from the given read implementation, decoding the bytes as
+    /// XDR, and construct the type implementing this interface from those
+    /// bytes.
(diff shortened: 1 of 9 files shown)


spec/output/generator_spec_rust/block_comments.x/MyXDR.rs
```diff
@@ -232,29 +232,53 @@ where
     #[cfg(feature = "std")]
     fn read_xdr(r: &mut impl Read) -> Result<Self>;
 
+    #[cfg(feature = "std")]
+    fn read_xdr_to_end(r: &mut impl Read) -> Result<Self> {
+        let s = Self::read_xdr(r)?;
+        // Check that any further reads, such as this read of one byte, read no
+        // data, indicating EOF. If a byte is read the data is invalid.
+        if r.read(&mut [0u8; 1])? == 0 {
```
jonjove, line 240: I'm not sure this matches up with the behavior of `std::io::Read::read_to_end`. Specifically, the [documentation](https://doc.rust-lang.org/std/io/trait.Read.html#method.read_to_end) reads

> This function will continuously call [read()](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) to append more data to buf until [read()](https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read) returns either [Ok(0)](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Ok) or an error of non-[ErrorKind::Interrupted](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Interrupted) kind.

but this does not handle `ErrorKind::Interrupted`. I've never worked with this trait before so I'm not really sure of the significance here.

leighmcculloch: Thanks, that's a good observation. I had read about the error and decided to treat it also as an error because I thought that's what the behavior of the function was, but I think I was reading the docs for `read` rather than `read_to_end` in that moment because a re-read yes definitely say it continues reading when interrupted. I'll update our implementation to match. I've reviewed the source for `read_to_end` as well now.

leighmcculloch: I've written tests for interrupted in https://github.com/stellar/rs-stellar-xdr/pull/178, and all read_xdr implementations continue on `Interrupted`, so this is handled by that. As part of this PR I'm adding docs to the read functions describing their behaviors, and I've added a note to all of the functions saying that they are required to internally continue on interrupted.

Issue #125, opened 2022-07-21
Rust: serde json encode VecM<u8> for XDR strings as strings


leighmcculloch:
The Rust generated code stores XDR strings as vectors of u8's because Rust's definition of a string is strictly UTF8, where-as the XDR definition is loose supporting any byte sequences.

However, for the serde JSON encoding of the XDR it would be ideal if the fields designated as strings could be rendered as Strings.

This raises the question: What should the serde serialization do if the vector of u8's contains non-UTF8 sequences? Should it:
  - Panic.
  - Convert lossy.

Alternatively: Is there a way in Rust we could render a string in the serde JSON output that is not UTF8 safe by reverting to displaying `\xAB` style characters. This would be preferable because it would preserve the encoding, while still allowing us to optimize for human friendly string rendering.

This issue is similar and in the same category of work as #118 and #119.

Issue #119, opened 2022-07-13
Rust: serde json encode [u8; N] as hex string


leighmcculloch:
### What
Change the serde json encoding of `[u8; N]` as a hex string in the Rust generated code.

### Why
It is less common than encoding binary as base64, but the vast majority of our fixed type u8's in the Stellar XDR represent small values that we often share in hex format, so this puts the encoding of XDR closer to the native form we would typically discuss the values.

If this presents significant challenges then I think simply encoding as base64, like #118, would also be okay.

Follow up to #112 #117

Issue #118, opened 2022-07-13
Rust: serde json encode VecM<u8> as base64


leighmcculloch:
### What
Change the serde json encoding of `VecM<u8>` as base64 in the Rust generated code.

### Why
It is common to transmit binary as base64 in JSON and most JSON tooling has good support for dealing with it in this format. e.g. `jq`.

Follow up to #112 #117

leighmcculloch:
In the PRs I'm implementing this I've gone with hex instead of base64. While I prefer base64, and in most of our tools we appear to favor base64 for variable length binary data, in Soroban we use variable length binary fields frequently for conceptually fixed length fields. This means more often than not we have things like hashes or IDs in the vararray fields. I think we'll benefit by using hex since we often look at these types of data items in hex. It's also not much of a downside since it is just as trivial to encode and decode hex as it is base64 in most places.

PR #143, opened 2022-10-12
Rust: Type Opaque as BytesM and String as StringM

leighmcculloch:
### What
Rust: Type Opaque as BytesM and String as StringM

### Why
The ultimate goal of this change is to get us to a place where we can improve the rendering of XDR Strings and XDR Opaque types in stringified renderings of XDR values either in Rust Debug form or in JSON.

Rust has a powerful type system, but also some limitations that haven't been worked out yet, and so because we put Strings, Opaques (Bytes), and arrays for types into the same Rust type, VecM, we cannot specialize their rendering. This results in the default rendering being used by whatever rendering system is rendering the type. Most rendering systems in Rust, like the default Debug implementations, and serde_json, all rely on using types to distinguish how the rendering occurs, or they require adding attributes to field attributes which only works for values encoded within another type and not values encoded just as themselves.

This is an experiment at this point, and I'm reasonably confident there are dragons to be found.

There's a risk that this is reasonably disruptive to existing code. Not in the sense of changing behavior, as the behavior will stay the same, but in terms of just lines changed and potential collissions with other work. So if we proceed with it, some care will have to be taken on when to make the upstream changes. And some care can be taken to weigh up if the benefits of this change really outweigh any costs we discover along the way.

Close https://github.com/stellar/xdrgen/issues/125
Close https://github.com/stellar/xdrgen/issues/119
Close https://github.com/stellar/xdrgen/issues/118

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -577,7 +577,7 @@ module Xdrgen
           out.puts %{#[cfg_attr(feature = "arbitrary", derive(Arbitrary))]}
           out.puts "#[derive(Default)]" if is_var_array_type(typedef.type)
           if is_fixed_array_opaque(typedef.type)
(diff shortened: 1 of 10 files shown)


graydon (approved):
LGTM

PR #147, opened 2022-11-30
Rust: Change JSON format from camelCase to snake_case

### LINKED ISSUE #145: Rust: Change JSON to snake case
Change JSON to snake case. 
```rust
#[serde(rename_all = "snake_case")]
```

It is currently camel case. But most of JSON in the Stellar ecosystem is snake case.

For example:
- All SEPs that provide APIs are snake case. 
- stellar.expert API is snake case ([ref](https://stellar.expert/openapi.html#tag/Payment-Locator-API/operation/findPayments))
- Horizon is snake case.
- Block Daemon API is snake case ([ref](https://ubiquity.docs.blockdaemon.com/))
- Lobstr Vault is snake case ([ref](https://github.com/Lobstrco/Vault-Android#developers-integrate-lobstr-vault-with-your-service))

The only Stellar API I could find that wasn't snake case was the stellarbeat.io API ([ref](https://api.stellarbeat.io/docs/#/Node/getNodeByPublicKey)).

Uniformity doesn't really matter, but if folks are consistently pick one over the other I don't think there's much advantage to picking something else. 

This will be breaking, but given the primary use of the generated Rust code is in Soroban which is accepting breaking changes that should be fine. We are making breaking changes to the XDR and subsequently the JSON frequently anyway. 

Alternatively we could make it a feature to choose the case, but I don't know of significant benefit of us doing that.

leighmcculloch:
### What
Change JSON format from camelCase to snake_case in the Rust generated output.

### Why
Close #145

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -100,7 +100,7 @@ module Xdrgen
         #[cfg_attr(
           all(feature = "serde", feature = "alloc"),
           derive(serde::Serialize, serde::Deserialize),
-          serde(rename_all = "camelCase")
+          serde(rename_all = "snake_case")
         )]
         pub enum TypeVariant {
             #{types.map { |t| "#{t}," }.join("\n")}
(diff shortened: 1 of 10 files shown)


tomerweller (approved):

leighmcculloch:
I've merged this to include this in the next Soroban release. But if a majority of folks are concerned I can roll it back trivially.

PR #155, opened 2023-03-22
Golang: add bool support to EncodeTo() and DecodeFrom()

### LINKED ISSUE #154: Go: bool Decode tries to deref a non-pointer
### What did you do?

See: https://github.com/stellar/go/pull/4815#discussion_r1140459745

2opremio:
Closes #154 

Also, for types we can't or don't know how to decode directly,  make sure we always call `Decode()` on a pointer reference

You can see the result of this PR at https://github.com/stellar/go/pull/4818


2opremio:
As mentioned before ( https://github.com/stellar/xdrgen/pull/135#discussion_r964412074 ) we should really include an automatic Go encode/decode roundtrip test on the generated files. This problem would had been detected (as shown by the changes in the generated files).

tsachiherman (approved):

PR #160, opened 2023-07-12
feat(java): render constants.

overcat:
This PR will render all constants into the `Constants` class.


overcat:
Hi @tamirms, can you review and merge this PR?

PR #165, opened 2023-08-10
Escape certain reserved terms when building member functions

### LINKED ISSUE #153: JS: Using `set` as an arm name for SCSpecTypeDef breaks the js-xdr package.
### What version are you using?

The latest commit on `next`, `df18148747e807618acf4639db41c4fd6f0be9fc`.

### What did you do?

I generate the JavaScript version of XDR using `xdrgen` which imports `js-xdr`. Because `js-xdr` uses the keyword `set` when it tries to interact with `SCSpecTypeDef` it breaks.

```
union.js:65 Uncaught (in promise) Error: set not set
    at ChildUnion2.get (union.js:65:15)
    at ChildUnion2.get [as set] (union.js:219:23)
    at ChildUnion2.Union2 (union.js:44:10)
    at new ChildUnion2 (union.js:146:111)
    at Function.read (union.js:120:14)
    at var-array.js:60:33
    at baseTimes (_baseTimes.js:15:21)
    at times (times.js:44:16)
    at VarArray2.read (var-array.js:59:26)
    at struct.js:57:26
```

The offending XDR is here.

```
union SCSpecTypeDef switch (SCSpecType type)
{
case SC_SPEC_TYPE_VAL:
case SC_SPEC_TYPE_U64:
case SC_SPEC_TYPE_I64:
case SC_SPEC_TYPE_U128:
case SC_SPEC_TYPE_I128:
case SC_SPEC_TYPE_U32:
case SC_SPEC_TYPE_I32:
case SC_SPEC_TYPE_BOOL:
case SC_SPEC_TYPE_SYMBOL:
case SC_SPEC_TYPE_BITSET:
case SC_SPEC_TYPE_STATUS:
case SC_SPEC_TYPE_BYTES:
case SC_SPEC_TYPE_ADDRESS:
    void;
case SC_SPEC_TYPE_OPTION:
    SCSpecTypeOption option;
case SC_SPEC_TYPE_RESULT:
    SCSpecTypeResult result;
case SC_SPEC_TYPE_VEC:
    SCSpecTypeVec vec;
case SC_SPEC_TYPE_MAP:
    SCSpecTypeMap map;
case SC_SPEC_TYPE_SET:
    SCSpecTypeSet set;
case SC_SPEC_TYPE_TUPLE:
    SCSpecTypeTuple tuple;
case SC_SPEC_TYPE_BYTES_N:
    SCSpecTypeBytesN bytesN;
case SC_SPEC_TYPE_UDT:
    SCSpecTypeUDT udt;
};
```

Specifically this line.

```
// ...
case SC_SPEC_TYPE_SET:
    SCSpecTypeSet set;
// ...
```

### What did you expect to see?

No errors when using `js-xdr` with the latest XDR.

### What did you see instead?

The `set not set` error.

We can fix this by simply uppercasing the `set` to `Set`, or picking another name.

Shaptic:
Closes #153.

Note: I'm still testing the end-to-end with js-stellar-base.

### DIFF
--- a/lib/xdrgen/generators/javascript.rb
+++ b/lib/xdrgen/generators/javascript.rb
@@ -131,8 +131,6 @@ module Xdrgen
                   if union.discriminant.type.is_a?(AST::Typespecs::Int)
                     member = union.resolved_case(acase)
                     "#{member.value}"
-                  elsif member_name(acase.value) == 'set'
-                    '"set_"'
                   else
                     '"' + member_name(acase.value) + '"'
                   end
@@ -192,7 +190,11 @@ module Xdrgen
       end
 
       def member_name(member)
-        name(member).camelize(:lower)
+        fixedName = name(member).camelize(:lower)
+        if fixedName == 'set'
+          return 'set_'
+        end
+        fixedName
       end
 
       def reference(type)
--- a/lib/xdrgen/generators/javascript.rb
+++ b/lib/xdrgen/generators/javascript.rb
@@ -131,6 +131,8 @@ module Xdrgen
                   if union.discriminant.type.is_a?(AST::Typespecs::Int)
                     member = union.resolved_case(acase)
                     "#{member.value}"
+                  elsif member_name(acase.value) == 'set'
+                    '"set_"'
                   else
                     '"' + member_name(acase.value) + '"'
                   end


tsachiherman (approved):
I think that at some point we'd need to extend that to all the [js reserved words](https://www.w3schools.com/js/js_reserved.asp), but at this point in time, the fix looks as if it would produce a good workaround.

Shaptic:
@tsachiherman it's not that `set` is reserved by JS (it isn't), it's that `set` is a method used internally by js-xdr to actually build the structures. Thus, having an _actual member_ named `set` overwrites that, making it impossible to, well, actually set fields.

tsachiherman:
> @tsachiherman it's not that `set` is reserved by JS (it isn't), it's that `set` is a method used internally by js-xdr to actually build the structures. Thus, having an _actual member_ named `set` overwrites that, making it impossible to, well, actually set fields.

so, you're saying that theoretically it would be a parameter provided to the xdrgen to rename a certain names ?

Shaptic:
@tsachiherman we can expand the list in `member_name` as needed of course, now, but I don't think it needs to be configurable if that's what you're implying

I'm just saying that the root cause is more-so that `set` is "reserved" by js-xdr internally rather than by the language itself.

PR #173, opened 2023-10-20
Allocate memory for vararray as needed in Rust

leighmcculloch:
### What
Allocate memory for vararray as needed in Rust generated code.


jayz22:
There are more places to fix. For example, some calling `vec![0; len]` https://github.com/stellar/xdrgen/blob/master/lib/xdrgen/generators/rust/src/types.rs#L1222

graydon (approved):

PR #178, opened 2023-10-28
Rust: Limit preallocation during byte (opaque, string) reads to 1kb

### LINKED ISSUE #174: Rust: Remove preallocation in vararray / varopaque conversions
Render varopaque conversions as buffered copies that allocate as the read occurs, instead of as a single allocate upfront.

This affects the three types `VecM`, `BytesN`, `String` that use this code:

```rust
Vec::with_capacity(...)
```

```rust
vec![0u8; len as usize]
```

leighmcculloch:
### What
In the Rust generated code limit preallocation during byte (opaque, string) reads to 1kb.

### Why
Close #174

### DIFF
--- a/lib/xdrgen/generators/rust/src/types.rs
+++ b/lib/xdrgen/generators/rust/src/types.rs
@@ -653,6 +653,35 @@ fn pad_len(len: usize) -> usize {
     (4 - (len % 4)) % 4
 }
 
+/// Read exactly the specified length of bytes from the read into a new Vec that
+/// is returned in batches. The function will preallocate only the memory
+/// required for each batch as it goes while reading so that no large
+/// preallocation occurs without the message data being available.
+#[cfg(feature = "std")]
+fn read_exact_in_batches<R: Read>(r: R, len: usize, batch_size: usize) -> Vec<u8> {
+    let mut vec = vec![0u8; 0];
+    let mut len_remaining = len as usize;
+
+    // Read one batch at a time.
+    while len_remaining > 0 {
+        let len_read = core::cmp::min(len_remaining, batch_size);
+        let offset = vec.len();
+        let len_new = vec.len() + len_read;
+
+        // Reserve exactly the new space required. Reserving exact prevents vec
+        // from increasing the allocated memory 2x which could result with large
+        // unnecessary allocations.
+        vec.reserve_exact(len_new);
+
+        // Resize the vec so that the read can read into the next batch of space.
+        vec.resize(len_new, 0);
+
+        r.read_exact(&mut vec[offset..])?;
+        len_remaining -= len_read;
+    }
+    vec
+}
+
 impl ReadXdr for i32 {
     #[cfg(feature = "std")]
     fn read_xdr<R: Read>(r: &mut DepthLimitedRead<R>) -> Result<Self> {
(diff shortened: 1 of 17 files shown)


lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -3,6 +3,9 @@ use core::{array::TryFromSliceError, fmt, fmt::Debug, marker::Sized, ops::Deref,
 #[cfg(feature = "std")]
 use core::marker::PhantomData;
 
+#[cfg(feature = "std")]
+const MAX_PREALLOCATED_BYTES_READ: usize = 1024; // 1KB
```
leighmcculloch, line 7: Would be nice to make this configurable. But would rather set it to a sensible limit at this point and revisit that when it matters.

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -1219,8 +1222,15 @@ impl<const MAX: u32> ReadXdr for VecM<u8, MAX> {
...
-            let mut vec = vec![0u8; len as usize];
-            r.read_exact(&mut vec)?;
+            let mut vec = vec![0u8; 0];
+            let mut len_remaining = len as usize;
+            while len_remaining > 0 {
+                let len_read = core::cmp::min(len_remaining, MAX_PREALLOCATED_BYTES_READ);
+                let offset = vec.len();
+                vec.resize(vec.len() + len_read, 0);
+                r.read_exact(&mut vec[offset..])?;
+                len_remaining -= len_read;
```
graydon, line 1232: the only thing that I'm slightly concerned with here is just underflow paranoia around "what if the arithmetic is wrong or read_exact misbehaves"; maybe .saturating_sub() or .checked_sub()?

graydon (changes requested):
One minor nit otherwise looks great

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -1219,8 +1222,15 @@ impl<const MAX: u32> ReadXdr for VecM<u8, MAX> {
                 return Err(Error::LengthExceedsMax);
             }
 
-            let mut vec = vec![0u8; len as usize];
-            r.read_exact(&mut vec)?;
+            let mut vec = vec![0u8; 0];
+            let mut len_remaining = len as usize;
+            while len_remaining > 0 {
```
dmkozh, line 1227: It seems like it should be easy to factor this out to a helper function. It's way too easy to mess up when updating the same code in three different places.

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -1219,8 +1222,15 @@ impl<const MAX: u32> ReadXdr for VecM<u8, MAX> {
...
+                let len_read = core::cmp::min(len_remaining, MAX_PREALLOCATED_BYTES_READ);
+                let offset = vec.len();
+                vec.resize(vec.len() + len_read, 0);
```
tsachiherman, line 1230: maybe I'm not following the goal here, but wouldn't these resize calls defeat the purpose of this PR ?

leighmcculloch: The resize calls rather than preallocating the entire buffer up front defer the allocation until we've been able to parse some of the input.

For example, for a 2048 byte input, the logic will allocate 1024 memory, and then parse the first 1024 bytes before allocating more memory.

lib/xdrgen/generators/rust/src/types.rs
leighmcculloch: Ah yes this is a really good observation, because read_exact is part of the trait and overridable rather than being a fixed library function. I'll change the logic.

leighmcculloch: Looked into this more, the read_exact function can't communicate about how much it has read. It returns `()` in the success case. So the implementor of read_exact has no control over the `len_read` and `len_remaining` values, so I don't think any underflow risk exists.

I mean, the read_exact can misbehave, the data read might be corrupt, but the arithmetic won't meaningful play a role in that.

graydon: yeah, I mean I see that it's as correct as my brain can convince itself of; I guess I am just trying to be extra paranoid and defensive anywhere "wrong arithmetic could be fatal" and here it'd be an infinite-read loop.but of course there are limits to healthy paranoia and if you think it's overkill I guess I won't block this.

leighmcculloch: Yup, but read_exact can't affect the values of len_read and len_remaining. The only thing affecting them is the code in that function, so if there's a bug there, we are 100% responsible for it, not he implementor of read_exact.

leighmcculloch:
Alternative implementation that I believe solves the same problem is available at #179.

leighmcculloch:
Closing in favor of #179.

PR #179, opened 2023-10-31
Rust: Add length limits

### LINKED ISSUE #174: Rust: Remove preallocation in vararray / varopaque conversions

leighmcculloch:
### What
Add length limits to the Rust generated code.

### How
Simplified the existing DepthLimitedRead/Write structs, combining them into one, and renaming them to Limited. Added the Limits type that captures the limits and made it a required input to functions that used to collect the depth limit.

### Why

To avoid preallocating memory when parsing will ultimately fail because the message being parsed doesn't contain sufficient data for the parse to succeed.

Close #174

Alternative to https://github.com/stellar/xdrgen/pull/178

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -166,21 +166,21 @@ module Xdrgen
 
             #[cfg(feature = "std")]
             #[allow(clippy::too_many_lines)]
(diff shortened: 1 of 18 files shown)


graydon (approved):
LGTM

PR #193, opened 2024-03-04
Generate schemars derives in Rust generator

leighmcculloch:
### What
Generate `schemars::JsonSchema` derives in Rust generator whenever the existing `serde` feature is enabled.

### Why
To support generating JSON schemas of the JSON that gets generated for the types.

Close https://github.com/stellar/rs-stellar-xdr/issues/348

### Todo

- [x] Test out the initial prototype
- [x] Check if the schema is valid (ask @willemneal for help in validating it with JS tooling)
- [x] Consider the best way to structure the features.
- [x] Generate custom hex/hash for any fixed or var array custom types
- [x] Tidy and chefs kiss
- [ ] ~Adopt @willemneal's no additional properties from https://github.com/stellar/rs-stellar-xdr/pull/351#discussion_r1538955048~ _(Edit: Reduces size, but results in same affect, so leaving out and we can add as an improvement.)_
- [ ] ~Include comments from .x files~ _(Edit: We can do this as a follow up.)_
- [ ] ~Add json schema meta titles for union arms. (like how @willemneal did in a visitor, but add them in the original schema for the type in the generated code)~ _(Edit: Deferred.)_

### Related

- https://github.com/stellar/rs-stellar-xdr/pull/347


lib/xdrgen/generators/rust.rb
```diff
@@ -94,6 +95,26 @@ def render_lib(out)
...
+                <<-EOS
+                $macro_to_call_back!(#{t}, $($context),*);
+                EOS
+                end.join("\n")}
+
+            }};
+        }
+        pub use _call_macro_with_each_type_#{@output.inputs_hash} as call_macro_with_each_type;
+        EOS
+      end
```
leighmcculloch, line 116: I forgot there's an issue with this that I need to tweak before merging this. I need to replace the inputs hash with a configurable name to pass in.

sisuresh (approved):

PR #196, opened 2024-05-15
Java: show the XDR source code in the javadoc

overcat:
Previously, the way we rendered comments prevented XDR definitions from being displayed in javadoc, in this PR we have added it.

See https://github.com/lightsail-network/java-stellar-sdk/pull/595


overcat:
Hi @JakeUrban, can you review the PR?

leighmcculloch (approved):

leighmcculloch:
Nice, thanks for fixing the comment rendering.

Some day it'd be awesome if it also took the comments in the XDR and made them available to the generators for inserting into the code. So that comments on an enum value, could be rendered by a generator onto the enum written into the code. And same for struct fields, structs themselves, etc.

PR #199, opened 2024-06-27
Java: throw a more appropriate exception.

overcat:
This PR is very simple, I think it's more appropriate to throw an `IllegalArgumentException` here.

### DIFF
--- a/lib/xdrgen/generators/java/XdrString.erb
+++ b/lib/xdrgen/generators/java/XdrString.erb
@@ -29,7 +29,7 @@ public class XdrString implements XdrElement {
   public static XdrString decode(XdrDataInputStream stream, int maxSize) throws IOException {
     int size = stream.readInt();
     if (size > maxSize) {
-      throw new InvalidClassException("String length " + size + " exceeds max size " + maxSize);
+      throw new IllegalArgumentException("String length " + size + " exceeds max size " + maxSize);
     }
     byte[] bytes = new byte[size];
     stream.read(bytes);
--- a/lib/xdrgen/generators/java.rb
+++ b/lib/xdrgen/generators/java.rb
@@ -218,7 +218,7 @@ module Xdrgen
         end
         out.puts <<-EOS.strip_heredoc
             default:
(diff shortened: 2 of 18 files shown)


lib/xdrgen/generators/java/XdrString.erb
```diff
@@ -29,7 +29,7 @@ public class XdrString implements XdrElement {
   public static XdrString decode(XdrDataInputStream stream, int maxSize) throws IOException {
     int size = stream.readInt();
     if (size > maxSize) {
-      throw new InvalidClassException("String length " + size + " exceeds max size " + maxSize);
+      throw new IllegalArgumentException("String length " + size + " exceeds max size " + maxSize);
```
leighmcculloch, line 32: Because it's the size inside the stream, and not the maxSize variable, that is illegal, I'm not sure I'd class this as an illegal argument exception (IAE) because it's the arguments are actually okay it's an issue with data available on the stream. [BufferOverflowException (BOE)](https://docs.oracle.com/javase/8/docs/api/java/nio/BufferOverflowException.html) might be more appropriate?

Happy to merge this if you think IAE is the most appropriate exception over BOE. Wdyt?

leighmcculloch (commented):
👍🏻 The change looks good to me. One question inline. Let me know and if you want to merge it as is happy to do that.

lib/xdrgen/generators/java/XdrString.erb
overcat: Hi @leighmcculloch, BOE is also a suitable choice, but I still prefer to use IAE here.

- In the given code, the exception is triggered when one parameter (size) exceeds another parameter (maxSize). This aligns well with the definition of IAE, which is used when a method receives an inappropriate or invalid argument. In this case, the size being greater than maxSize is due to an incorrect size parameter contained within the passed-in stream.
- IAE is a very common and familiar exception type in Java, widely used to represent various invalid arguments or inappropriate conditions. By using IAE, other developers can quickly understand and handle the exception.

PR #202, opened 2024-07-30
Rust: Enum names not only numbers

leighmcculloch:
### What
When generating rust code for enums that when their shared prefix is removed start with a number, add a letter to the start of the enum member name.

### Why
Enums that share a prefix followed by a number result in invalid Rust code being generated. Rust, like most languages, does not allow identifiers to begin with a number.

For example:
```c
enum BinaryFuseFilterType
{
    BINARY_FUSE_FILTER_8_BIT = 0,
    BINARY_FUSE_FILTER_16_BIT = 1,
    BINARY_FUSE_FILTER_32_BIT = 2
};
```

#### Before
```rust
pub enum BinaryFuseFilterType {
  8Bit = 0,
  16Bit = 1,
  32Bit = 2,
}
```

#### After
```rust
pub enum BinaryFuseFilterType {
  B8Bit = 0,
  B16Bit = 1,
  B32Bit = 2,
}
```


graydon (approved):
Not going to pretend I speak Ruby but this looks fairly harmless

PR #203, opened 2024-08-28
Rust: Add fn for deserializing multiple types in json from a stream

leighmcculloch:
### What
Add deserialize_json fn to the Type object for deserializing multiple types in json from a stream when the type variant is known at runtime not compile time. Also rename read_json to from_json in a backwards compatible way.

### Why
In the CLI we need to be able to decode JSON for arbitrary types at runtime in much the same way we can decode XDR for arbitrary types so the commands have symmetry.

We support decoding XDR to JSON for streams, but not in reverse.

There's a function today `read_json` that can decode a JSON object into an arbitrary Type variant, but because of the interface the function used it only does so for a single JSON object and expects an EOF to occur.

For streams we need to use the more complex serde interface to deserialize multiple objects from the stream.

The read_json fn is renamed to from_json to be more similar to the serde_json's own naming schemes for these two functions. It's much easier to see the correlation that the from method in our crate calls a from method on the serde_json crate which has certain meaning, vs the deserialize call which has certain meaning in that crate too.

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -250,10 +250,24 @@ module Xdrgen
             }
 
             #[cfg(all(feature = "std", feature = "serde_json"))]
-            #[allow(clippy::too_many_lines)]
+            #[deprecated(note = "use from_json")]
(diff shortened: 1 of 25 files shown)


sisuresh (approved):
LGTM

PR #205, opened 2024-09-12
Rust: add support for no_std env

overcat:
Enable rs-stellar-xdr to encode/decode XDR in a `no_std` environment.

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -175,7 +175,7 @@ module Xdrgen
             pub const VARIANTS: [TypeVariant; #{types.count}] = [ #{types.map { |t| "TypeVariant::#{t}," }.join("\n")} ];
             pub const VARIANTS_STR: [&'static str; #{types.count}] = [ #{types.map { |t| "\"#{t}\"," }.join("\n")} ];
 
-            #[cfg(feature = "std")]
+            #[cfg(feature = "alloc")]
             #[allow(clippy::too_many_lines)]
             pub fn read_xdr<R: Read>(v: TypeVariant, r: &mut Limited<R>) -> Result<Self> {
                 match v {
@@ -190,7 +190,7 @@ module Xdrgen
                 Ok(t)
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(feature = "alloc")]
             pub fn read_xdr_to_end<R: Read>(v: TypeVariant, r: &mut Limited<R>) -> Result<Self> {
                 let s = Self::read_xdr(v, r)?;
                 // Check that any further reads, such as this read of one byte, read no
@@ -234,7 +234,7 @@ module Xdrgen
                 }
             }
 
-            #[cfg(feature = "std")]
+            #[cfg(feature = "alloc")]
             pub fn from_xdr<B: AsRef<[u8]>>(v: TypeVariant, bytes: B, limits: Limits) -> Result<Self> {
                 let mut cursor = Limited::new(Cursor::new(bytes.as_ref()), limits);
                 let t = Self::read_xdr_to_end(v, &mut cursor)?;
@@ -318,7 +318,7 @@ module Xdrgen
         }
 
         impl WriteXdr for Type {
-            #[cfg(feature = "std")]
+            #[cfg(feature = "alloc")]
             #[allow(clippy::too_many_lines)]
             fn write_xdr<W: Write>(&self, w: &mut Limited<W>) -> Result<()> {
                 match self {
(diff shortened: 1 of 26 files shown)


lib/xdrgen/generators/rust.rb
```diff
@@ -76,7 +76,7 @@ def render_top_matter(out)
           //  #{@output.relative_source_paths.join("\n//  ")}
         EOS
         out.break
-        out.puts "#![allow(clippy::missing_errors_doc, clippy::unreadable_literal)]"
+        out.puts "#![allow(clippy::missing_errors_doc, clippy::unreadable_literal, clippy::needless_question_mark)]"
```
overcat, line 79: I did not modify the relevant code, so I am not sure why clippy is giving a warning. And I encountered some difficulties in fixing this warning, so let's add it first.

leighmcculloch:
Thanks for looking into this.

Why `embedded-io` over `core_io`? Are there other crates to consider? I haven't looked into this enough yet to know myself, but `core_io` is more of a mirror of `std::io`, and it has a `Cursor` type.

I noticed that we're using `embedded_io_extras` to get a `Cursor` type. Do you think that could be contributed upstream to reduce the nested dependencies we're adding to the xdr lib? I also wonder if embedded-hal has an equivalent type we're missing.

Rather than add a generic sounding 'alloc' feature to the stellar-xdr crate, I'm inclined to instead add a feature named after the optional dependency. So if you enable the embedded-io feature, you get embedded-io capability.

We need to think about how to test this for consistency with the std::io behavior. The xdr lib depends on the std::io::Read::read_exact function, and it's default_read_exact implementation, and it has some subtle interactions with reading EOF safeties in the xdr lib. But I notice that the embedded-io read_exact function isn't identical, and so I think we'll be hesitant to merge changes unless we can either avoid the difference, or do some differential fuzzing to ensure there's no breakage or observable difference in behavior.

overcat:
Hi @leighmcculloch, thank you for taking the time to review the PR.

> Why `embedded-io` over `core_io`? Are there other crates to consider? I haven't looked into this enough yet to know myself, but `core_io` is more of a mirror of `std::io`, and it has a `Cursor` type.

I researched the following libraries:
- [ciborium_io](https://crates.io/crates/ciborium-io): Only provides `Read/Write` traits and differs significantly from `std::io::Read/Write`.
- [core2](https://crates.io/crates/core2): Offers an API almost identical to `std::io`, but hasn't been maintained for over two years. I'm concerned about the project's future. #204 includes an implementation using `core2`.
- [core_io](https://crates.io/crates/core_io): Last released in 2021 with very few downloads.
- [embedded-io](https://crates.io/crates/embedded-io): Maintained by rust-embedded, actively developed. I believe it's currently the best choice.

> I noticed that we're using embedded_io_extras to get a Cursor type. Do you think that could be contributed upstream to reduce the nested dependencies we're adding to the xdr lib? I also wonder if embedded-hal has an equivalent type we're missing.

I will create an issue to ask if they would consider adding `Cursor` to `embedded-io`. Another option is to implement `Cursor` in `rs-stellar-xdr`, allowing us to depend directly on `embedded-io`. I think `Cursor` does not exist in `embedded-hal`.


> Rather than add a generic sounding 'alloc' feature to the stellar-xdr crate, I'm inclined to instead add a feature named after the optional dependency. So if you enable the embedded-io feature, you get embedded-io capability.

I like this idea, so that embedded-io will only be used when the user explicitly enables it.

> We need to think about how to test this for consistency with the std::io behavior. The xdr lib depends on the std::io::Read::read_exact function, and it's default_read_exact implementation, and it has some subtle interactions with reading EOF safeties in the xdr lib. But I notice that the embedded-io read_exact function isn't identical, and so I think we'll be hesitant to merge changes unless we can either avoid the difference, or do some differential fuzzing to ensure there's no breakage or observable difference in behavior.

I need to do a bit more research. If I have any questions, can I reach out to you on Discord?

overcat:
https://github.com/rust-embedded/embedded-hal/issues/631

overcat:
I added some tests to verify whether the behavior of embedde io meets expectations.

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -2767,3 +2816,335 @@ mod test {
         assert_eq!(v.to_option(), Some(1));
     }
 }
+
+#[cfg(all(test, any(feature = "std", feature = "embedded_io")))]
```
overcat, line 2820: I added some tests to verify whether the behavior of embedde io meets expectations.

PR #206, opened 2025-02-04
chore: pin 'concurrent-ruby' to a working version.

overcat:
See https://stackoverflow.com/questions/79360526/uninitialized-constant-activesupportloggerthreadsafelevellogger-nameerror

Running xdrgen will result in the following error, and this PR will fix the issue.
```
/usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support/logger_thread_safe_level.rb:16:in `<module:LoggerThreadSafeLevel>': uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger (NameError)

    Logger::Severity.constants.each do |severity|
          ^^^^^^^^^^
        from /usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support/logger_thread_safe_level.rb:9:in `<module:ActiveSupport>'
        from /usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support/logger_thread_safe_level.rb:8:in `<top (required)>'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from /usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support/logger_silence.rb:5:in `<top (required)>'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from /usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support/logger.rb:3:in `<top (required)>'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from /usr/local/rvm/gems/default/gems/activesupport-6.1.6.1/lib/active_support.rb:29:in `<top (required)>'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
        from <internal:/usr/local/lib/ruby/3.2.0/rubygems/core_ext/kernel_require.rb>:86:in `require'
[...]

PR #208, opened 2025-02-05
Fix xdrgen build breakage.

dmkozh:
There was an issue introduced in the new version of `concurrent-ruby` that breaks `active-support`. It seems to be easily resolved by just requiring logger before the affected library.


leighmcculloch:
Thanks for opening a fix. This is a duplicate of the following change that is merged now, so probably should be resolved.
- https://github.com/stellar/xdrgen/pull/206

PR #213, opened 2025-05-07
Rust: Serialize 64-bit integers as strings in xdr-json

### LINKED ISSUE #181: xdrjson: Rust: Render 64-bit integers as strings
### What
Render 64-bit integers as strings in JSON.

### Why
JS runtimes cannot correctly decode JSON numbers greater than 53-bits.

leighmcculloch:
### What
Add handling for 64-bit integers in XDR structures to ensure they are serialized as strings in JSON. Adds the handling in the cases of struct fields, union (enum) fields, and typedef (unamed struct) fields.

### Why
JavaScript has limited precision for integers larger than 53 bits, requiring large integers to be represented as strings to prevent data loss. Today those integers are rendered as numbers which requires that folks use non-native JSON parsers in JavaScript runtimes.

Close #181

### DIFF
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -1029,7 +1029,7 @@ module Xdrgen
         base_ref = base_reference(type)
         if ['i64','u64'].include?(base_ref)
           ref = reference(parent, type, 'DisplayFromStr')
-          "#[serde_as(as = \"#{ref}\")]"
+          "#[cfg_attr(all(feature = \"serde\", feature = \"alloc\"), serde_as(as = \"#{ref}\"))]"
         else
           nil
         end
--- a/lib/xdrgen/generators/rust.rb
+++ b/lib/xdrgen/generators/rust.rb
@@ -1025,7 +1025,7 @@ module Xdrgen
       def field_attrs(parent, type)
         base_ref = base_reference(type)
         if ['i64','u64'].include?(base_ref)
-          ref = reference(parent, type, 'serde_with::DisplayFromStr')
+          ref = reference(parent, type, 'DisplayFromStr')
           "#[serde_as(as = \"#{ref}\")]"
         else
           nil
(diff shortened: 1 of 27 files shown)


lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -154,9 +159,6 @@ impl From<Error> for () {
     fn from(_: Error) {}
 }
 
-#[allow(dead_code)]
-type Result<T> = core::result::Result<T, Error>;
```
leighmcculloch, line 158: The type alias is removed because the serde_as crate assumes `Result` in the current namespace is the `core::result::Result`.

lib/xdrgen/generators/rust/src/types.rs
```diff
@@ -914,6 +916,55 @@ impl<T: schemars::JsonSchema, const MAX: u32> schemars::JsonSchema for VecM<T, M
     }
 }
 
+#[cfg(feature = "schemars")]
+impl<T, TA, const MAX: u32> serde_with::schemars_0_8::JsonSchemaAs<VecM<T, MAX>> for VecM<TA, MAX>
```
leighmcculloch, line 920: This interface is required so that VecM<> can be referenced in serde_as attributes.

lib/xdrgen/generators/rust.rb
```diff
@@ -388,18 +388,20 @@ def render_source_comment(out, defn)
 
       def render_struct(out, struct)
         out.puts "#[derive(Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]"
+        out.puts %{#[cfg_eval::cfg_eval]}
```
leighmcculloch, line 391: `cfg_eval` is added on each type that uses serde_as field attributes, to add support for using cfg_attr on those field attributes.

dmkozh (approved):

PR #220, opened 2025-05-16
Add support for custom XDR generators

leighmcculloch:
### What
  Add support for providing custom generators to the XDR compilation process.

  ### Why
  To make it possible to use the xdrgen library to generate xdr with generators not defined in the library. As a first step for moving the generators to their respective repos. Some generators, such as the Rust generator and the JS generator, are tightly coupled with the library they generate code for. Because of that the separation of the generator into the xdrgen repo makes for an awkward development cycle, and since the code generated is not particularly portable, little benefit.

Related:
- https://github.com/stellar/stellar-protocol/discussions/1738
- https://github.com/stellar/rs-stellar-xdr/pull/449
- https://github.com/stellar/xdrgen/pull/221

### DIFF
--- a/lib/xdrgen/compilation.rb
+++ b/lib/xdrgen/compilation.rb
@@ -2,11 +2,12 @@ module Xdrgen
   class Compilation
     extend Memoist
 
-    def initialize(source_paths, output_dir:".", language: :ruby, namespace: nil, options: {})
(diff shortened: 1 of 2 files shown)


Shaptic (approved):
Great idea, :+1: :shipit:

dmkozh (approved):

